PR #4, opened 2017-12-19
Use gunagala to create imager array

### LINKED ISSUE #3: Fix create_imager_array
This has bee moved to AstroHuntsman/gunagala so we need to call from there.

wtgee:
Closes #3

### DIFF
--- a/huntsman/observatory.py
+++ b/huntsman/observatory.py
@@ -15,6 +15,8 @@ from pocs.utils import images as img_utils
 from huntsman.guide.bisque import Guide
 from huntsman.scheduler.observation import DitheredObservation
 
+from gunagala import imager
+
 
 class HuntsmanObservatory(Observatory):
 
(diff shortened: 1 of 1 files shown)


huntsman/observatory.py
```diff
@@ -27,7 +29,7 @@ def __init__(self, hdr_mode=False, with_autoguider=True, *args, **kwargs):
         # Creating an imager array object
         if self.has_hdr_mode:
             self.logger.info('\tSetting up HDR imager array')
-            self.imager_array = utils.hdr.create_imager_array()
+            self.imager_array = imager.create_imagers()
```
AnthonyHorton: Yes, I think this is right. `utils.hdr.create_imager_array()` dates from a long time ago. That was part of the code that was first moved into `utils.signal-to-noise.py`, and then into `gunagala`.

https://github.com/AstroHuntsman/HuntsmanPOCSLegacy/commit/3608dde0f21e5bea031aaa14938a91067724da68#diff-0cfce9b41a8fade31d2f4e528512652d

PR #28, opened 2018-01-03
Hue Lights

Physarah:
Ready for pull request after tested at the dome and remotely with factory reset recovery


huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
+import qhue
+from os import path
+from qhue import Bridge, QhueException, create_new_username
```
wtgee: You do a full `import qhue` above and use, for example, `qhu.create_new_username` below, so this `from qhue import ...` isn't used. You want one or the other.

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
...
+
+#note, light 5 can be 'uncommented' once installed, or causes problems using currently 
+
+
+#deal with error system - should errors be sent to pocs? how does pocs handle error messages?
+
+def manual_login_phillips_hue(hue_ip="10.88.21.10", verbose = False):
+    
+    """
```
wtgee: Same pep8 and docstring comments as in #29

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
+import qhue
+from os import path
+from qhue import Bridge, QhueException, create_new_username
+
+#note, light 5 can be 'uncommented' once installed, or causes problems using currently 
+
+
+#deal with error system - should errors be sent to pocs? how does pocs handle error messages?
+
+def manual_login_phillips_hue(hue_ip="10.88.21.10", verbose = False):
```
wtgee: I would rename file to `hue_light_control.py` and then drop the `phillips_hue` from all the function names.

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
+import qhue
+from os import path
+from qhue import Bridge, QhueException, create_new_username
+
+#note, light 5 can be 'uncommented' once installed, or causes problems using currently 
+
+
+#deal with error system - should errors be sent to pocs? how does pocs handle error messages?
```
wtgee: Which error messages are you referring to?

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
...
+    
+    if verbose:
+    
+       print("The username created is", username)
+    
+    b = qhue.Bridge(hue_ip,username)
+    
+    return(b)
+    
+def automatic_login_phillips_hue(hue_ip = "10.88.21.10", file_path = "hue_username.txt", verbose = False):
```
wtgee: We'll need the yaml config here too.

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
...
+        
+    Outputs:
+        
+        b: bridge connection
+        
+    """
+    
+    if not path.exists(file_path):
+        
+        while True:
```
wtgee: This `while` loop is just going to repeat the "Cannot create new username" message over and over. Is the idea that it does that until someone presses the button? If so there is no way to break from this if no one can press the button. We'll want to rethink this based on what is wanted, but I would just remove the `while` for now and `return` in the `except` since the whole automatic login cannot be performed.

huntsman/utils/light_control.py
```diff
@@ -0,0 +1,265 @@
...
+            try:
+                username = create_new_username(hue_ip)
+                break
+            except QhueException as err:
+                print("Cannot create new username: {}".format(err))
+                  
+        with open(file_path, "w") as cred_file:
```
wtgee: There should be one function `login` that both the `manual` and `automatic` use. In fact, the `manual` is not really needed. The one `login` should do `create_new_username` and write it to a file. The `automatic` should look for that file and use it otherwise call `login`.

huntsman/utils/light_control.py
Physarah: Fixed

huntsman/utils/light_control.py
Physarah: Fixed

huntsman/utils/light_control.py
Physarah: I used Autopep8 in spyder and it seemed to work

huntsman/utils/light_control.py
Physarah: Lee commented in the first pr that if a username can't be found and the bridge button needs to be pushed, instead of just giving the user the error it should be sent to pocs? But I can't seem to find the other pr comments after Astrohuntsman/POCS was changed.

huntsman/utils/light_control.py
Physarah: I think we might be able to get rid of the manual login all together, as the auto login really does this function already. Manual was more for me before the other function was working properly, and I kept it incase it was needed later.

huntsman/utils/light_control.py
wtgee: > I used Autopep8 in spyder and it seemed to work

You did this before and this is how the code looks, or you have since done this? If you had already done then something is not correct about it as there are many pep8 errors.

Physarah: I've since done this, so should be corrected now for the next pr

huntsman/utils/light_control.py
Physarah: Is there a yaml file for this I can play with or shall I leave the yaml config until it's been finalised?

wtgee: You will want to create something like `conf_files/lights.yaml`. Then you can do:

```
from pocs.utils.config import load_config

config_fn = '{}/conf_files/{}'.format(os.getenv('HUNTSMAN_POCS'), 'lights.yaml')
config = load_config(config_files=[config_fn])
```

This assumes you have the $HUNTSMAN_POCS environment variable set.

This also is obviously just using a separate config file from the others, which is fine for now. We might combine them or keep it separate.

huntsman/utils/hue_light_control.py
```diff
@@ -211,13 +211,28 @@ def hue_observing_mode(bridge, led_index, desk_index, verbose=False):
...
  
     
+   
+    
+    
+    
     
 class Boo:
     
     def hello(x, y):
```
wtgee: Should probably experiment with this stuff in another branch or somewhere else. ;)

Also, the first param to all methods inside a class should be `self`

PR #29, opened 2018-01-03
Acroname USBhub python interface

Physarah:
Tested at the dome and working, but yet to be tested remotely. Waiting for finalised YAML configuration files to integrate into huntsman-pocs


huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
+#A Python interface for the acroname USB Hubs - tested and functional
```
wtgee: This file needs some pep8 linting run on it, which will make it conform to style best-practices. If you are using Atom you can use this [package](https://atom.io/packages/pep8). Also see some of our text in the [CONTRIBUTING](https://github.com/panoptes/POCS/blob/e390b9990406f586d2be40b01ae1d7c5b86dd53d/CONTRIBUTING.md#code-formatting) doc.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        test: If true, will flash user LED's to verify that a connection has 
+        been made 
+        
+    Outputs: 
+        
+        stem: Connection to hub    
+    
+    """
+    
+    print('Connecting to hub', hub_serial)
```
wtgee: We might not always want to `print` so we need to control printing via a `verbose` flag. See an example [here](https://github.com/panoptes/POCS/blob/develop/pocs/utils/images/__init__.py#L194).

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    result = stem.discoverAndConnect(brainstem.link.Spec.USB, hub_serial)
+
+    if  result == (Result.NO_ERROR):
+        
+        result1 = stem.system.getSerialNumber()
+        
+        print ('Connected to USBStem with serial number:', result1)
+    
+        stem.system.setLED(1)
```
wtgee: What is the magic number `1`?

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+def connect_to_hub(hub_serial, verbose = False, test = True):
+    
+    """
+    Function: Used to connect to the hub and create stem
+    
+    Inputs:
+        
+        hub_serial: Serial number of the hub
+        
+        test: If true, will flash user LED's to verify that a connection has
```
wtgee: The name `test` is a little ambiguous for what this does. Something explicit like `toggle_leds` would be better.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    if  result == (Result.NO_ERROR):
+        
+        result1 = stem.system.getSerialNumber()
+        
+        print ('Connected to USBStem with serial number:', result1)
+    
+        stem.system.setLED(1)
+        
+        if test == True:
```
wtgee: You can just do `if test` otherwise the preferred is `if test is True`. (See note about renaming `test`)

huntsman/utils/hub_utils.py
wtgee: I see you have `verbose` as an option up above. For odd reasons I can explain in person it's better to use the `kwargs` as in example in comment above. Basically this allows `verbose` to pass through to other util functions without being "consumed" by this one.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            
+            if verbose: True
```
wtgee: This line does nothing. I think you were trying to add the verbose in which case it would be `if verbose is True`. But see `print` example above for way to avoid having to constantly re-write this.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            
+            print('Flashing the user LED on device')
+    
+            for i in range(0,101):
+        
+               flash = stem.system.setLED(i % 2)
```
wtgee: Not quite sure what is going on her although I see you are trying to flash. Should the `if` statement below be inside this `for` loop? Also, what is the logic behind this?

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+              sleep(0.1)  
+    else:
+        
+         print('Could not find hub with serial number:', hub_serial)
+         
+        
+    return(stem)
+    
+def yaml_import(no_camera = 0, yaml_file = "/Users/SC09015/Desktop/Astro/Code/device_info_testing_file.yaml"):
```
wtgee: As you mention, this will probably all change.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+"""        
+            
+    
+def disconnect_from_hub(stem, hub_serial, verbose = False):
+
+    stem.disconnect() 
+    
+    print('Disconnected from hub', hub_serial)
+    
+    return()
```
wtgee: No need to `return` a nothing.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    return(voltage)
+    
+
+def get_port_current(stem, port):
+    
+    current = stem.usb.getPortCurrent(port)
+     
+    return(current)
+    
+#remember stem = stem_serial if this doesnt work
```
wtgee: Not sure what this comment means or where it applies. Should be moved.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+def port_enable(stem, port, test = False):
+    
+    stem.usb.setPortEnable(port)
+    
+    if test == True:
+        
+       for i in range(1, 11):
+    
+          stem.system.setLED(i % 2)
+          time.sleep(2)
```
wtgee: This will end up sleeping for a total of 20 seconds while this flashes. Is that what you want when enabling a port?

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+#Test main for bug fixes - tested on site, all working
+          
+def main():
```
wtgee: Where is this called?

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    hub_serial = 0xCD12637D
```
wtgee: This will need to come from the YAML config when that is figured out.

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    stem = connect_to_hub(hub_serial)
+     
+    #Connect to hub1 port 0, turn on and get current and voltage reading turn off
+    #and disconnect
+    
+    port_enable(stem, port, test = True)
+    
+    current = get_port_current(stem, port)
+    print("The current output through port", port, "is", current.value)
```
wtgee: Note that you can use format strings to make this slightly better: 

```
print("The current output through port {} is {}".format(port, current.value))
```

huntsman/utils/hub_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    voltage = get_port_voltage(stem, port)
+    print("The voltage output from port", port, "is", voltage.value)
+    
+    port_disable(stem, port)
+    
+    disconnect_from_hub(stem, hub_serial)
+    
+    return()
```
wtgee: `return` not required.

wtgee (changes requested):
We'll want to wait to merge this until we figure out the yaml-config items but go through these comments and fix up, especially the pep8 style and the docstrings.

huntsman/utils/hub_utils.py
Physarah: Fixed similarly to hue lights

huntsman/utils/hub_utils.py
Physarah: 1 is LED on and 0 is LED off

huntsman/utils/hub_utils.py
Physarah: I forgot to remove this when I was testing packages for other model acronames. Removed now.

huntsman/utils/hub_utils.py
wtgee: I've never actually used the newish [`enum`](https://docs.python.org/3/library/enum.html) module but this would be a good case for it.

```
from enum import Enum

class Lights(Enum):
    OFF = 0
    ON = 1

# later on...
stem.system.setLED(Lights.ON)
```

huntsman/utils/hub_utils.py
Physarah: This could be removed now, if was just for testing purposes at the dome of the main functions.

huntsman/utils/hub_utils.py
Physarah: setLED can be 0 or 1, with 1 being LED on, and 0 LED off. So it's just cycling through 0 and 1 so that it flashes using the modulo.

wtgee:
@Physarah I'm surprised the autopep8 doesn't clean up the blank lines. pep8 says blank lines should be used ["sparingly"](https://www.python.org/dev/peps/pep-0008/#blank-lines) but you've got them for every other line.

I tend to view things like paragraphs when writing, which is to say that you want to group related ideas together without a space and then start a new "paragraph" with a blank line separating it from the other code.

Most things here are pretty small so are one "paragraph" each, but I would do, for instance:

```
    with open(yaml_file, 'r') as yml:
        try:
            newyml = (yaml.load(yml))
            hub_serial = newyml['cameras'][1]['devices'][no_camera]['USB_hub_serial_number']
            port = newyml['cameras'][1]['devices'][no_camera]['camera_into_serial_adaptor_port']
        except yaml.YAMLError as exc:
            print(exc)  # standard yaml error

    return(hub_serial, port)
```

instead of 
```
    with open(yaml_file, 'r') as yml:

        try:
            newyml = (yaml.load(yml))

            hub_serial = newyml['cameras'][1]['devices'][no_camera]['USB_hub_serial_number']

            port = newyml['cameras'][1]['devices'][no_camera]['camera_into_serial_adaptor_port']

        except yaml.YAMLError as exc:
            print(exc)  # standard yaml error

    return(hub_serial, port)
```

This is all one related piece of code except for the `return`, which is a bit separate. It's obviously kind of subjective but I would tend toward less blank lines, not more.

huntsman/utils/hub_utils.py
Physarah: I've removed the whole test function as it won't be used again now, but this is definitely useful for future reference!

huntsman/utils/hub_utils.py
Physarah: Fixed

requirements.txt
```diff
@@ -27,3 +27,5 @@ healpy
 astroquery
 photutils
 git+git://github.com/AstroHuntsman/gunagala@master#egg=gunagala
+brainstem >= 2.4
+yaml
```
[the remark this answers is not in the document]
Physarah: Fixed

huntsman/utils/hub_utils.py
Physarah: Removed, as you say, not required here.

huntsman/utils/hub_utils.py
Physarah: Fixed

huntsman/utils/hub_utils.py
Physarah: I like this! Implemented now.

AnthonyHorton: @wtgee There's an `enum` module? There are so many `dict`s & `list`s standing in for C enums in the SBIG, Birger & FLI interface code! This would be so much more neater.

wtgee: It's new in python 3.4 so it wasn't there when we started. Nice to finally have and easy to use. It's part of standard library too.

lspitler:
closing as not needed anymore.

PR #31, opened 2018-01-15
Creates a .yaml config file

moniquecockram:
Open up a new Jupyter Notebook in the same directory as the “generating_yaml.py” file is saved.
Type and run: `import generating_yaml`.
Use the drop down menus to choose the device information for one device set.
Once one device set’s information has been chosen, click ‘add new device set’
Then choose the information for the next device set.
Click ‘add new device set’ as above.
Repeat the above steps until all devices have been added.
Click ‘save file’ to generate the .yaml file in the two chosen locations.

### DIFF
--- a/scripts/generating_yaml.py
+++ b/scripts/generating_yaml.py
@@ -1,0 +1,213 @@
+
+# coding: utf-8
+
+# In[1]:
+
+from __future__ import print_function
+import yaml
+from ipywidgets import interact, interactive, fixed, interact_manual
+import ipywidgets as widgets
+from IPython.display import display
+import os
+import datetime
+
+
+# In[2]:
+
+device_data = {'serial_into_USBhub_port' : [0, 1, 2, 3, 4, 5, 6, 7],
+               'camera_into_serial_port' : [1, 2, 3, 4],
+               'USBhub_SN' : ['OX518EFFE1', 'OXCD12637D'],
+               'camera_into_USBhub_port' : [0, 1, 2, 3, 4, 5, 6, 7],
+               'sbig_SN' : ['83F011167', '83F011791', '83F011639', '83F010801', '83F010774', '83F011758', '83F010771', '83F011810'],
+              'birger_SN': ['10858', '14287', '14286', '14285', '13281', '13134', '14284', '13208', '14276'],
+              'lens_SN': ['3360000099', '3360000063', '3360000087', '2850000067', '3150000110', '5370000054'],
+              'filter_ID': ['r2_1', 'r2_2', 'g2_3', 'g2_4', 'r2_5', 'r2_6', 'r2_7', 'g2_8', 'g2_9', 'ha1_10', 'ha1_11', 'g2_12', 'ha1_13', 'ha1_14']
+              }
+
+
+# In[3]:
+
+with open('serial_management.yaml', 'w') as file:
+    yaml.dump(device_data)
+    file.write(yaml.dump(device_data))
+
+with open('serial_management.yaml', 'r') as file:
+    data = yaml.load(file)
+
+
+# In[4]:
+
+def type_filename(type_filename_here):
+    return type_filename_here
+
+
+# In[5]:
+
+file_name = interactive(type_filename, type_filename_here="filename.yaml")
+
+
+# In[6]:
+
+data_list = {'name' : 'HuntsmanSSO',
+             'log_file' : 'huntsman',
+             'location' : {'name' : 'Siding Spring Observatory',
+                           'latitude' : -31.27, #degrees
+                           'longitude': 149.06, # Degrees
+                          'elevation': 1000.0, # Meters
+                           'utc_offset': 10.00, # Hours
+                           'horizon': 30, # Degrees
+                           'timezone': 'Australia/Sydney'
+                          },
+             'scheduler' : {'type' : 'dispatch',
+                           'fields_file' : 'targets.yaml'},
+             'dome' : {'template_dir' : 'resources/bisque'},
+             'guider' : {'template_dir' : 'resources/bisque/guider',
+                         'image_path': '/tmp/guide_image.fits',
+                        'bin_size' : 1},
+             'mount' : {'brand': 'bisque',
+                       'model' : 45,
+                       'driver' : 'bisque',
+                       'template_dir' : 'resources/bisque'},
+             'pointing' : {'threshold' : 0.05,
+                          'exp_time' : 30,
+                          'max_iterations' : 3},
+             'state_machine' : 'huntsman',
+             'directories' : {'base' : '/var/huntsman', 'images' : 'images',
+                              'webcam' : 'webcams', 'data' : 'data',
+                              'resources' : 'POCS/resources/',
+                              'targets' : '/var/huntsman/huntsman-pocs/conf_files',
+                              'mounts' : 'POCS/resources/mounts'
+             },
+    'cameras' : {'hdr_mode' : True,
+               'auto_detect' : False,
+                'devices' : [None]}}
+
+
+# In[7]:
+
+def a(birger_serial_number):
+    return birger_serial_number
+
+def b(sbig_serial_number):
+    return sbig_serial_number
+
+def c(lens_serial_number):
+    return lens_serial_number
+
+def d(filter_ID_code):
+    return filter_ID_code
+
+def e(serial_into_USBhub_port):
+    return serial_into_USBhub_port
+
+def f(camera_into_serial_port):
+    return camera_into_serial_port
+
+def j(USBhub_SN):
+    return USBhub_SN
+
+def k(camera_into_USBhub_port):
+    return camera_into_USBhub_port
+
+
+# In[8]:
+
+birger_sn = data['birger_SN']
+birger_serial_number = interactive(a, birger_serial_number=birger_sn);
+sbig_sn = data['sbig_SN']
+sbig_serial_number = interactive(b, sbig_serial_number=sbig_sn);
+lens_sn = data['lens_SN']
+lens_serial_number = interactive(c, lens_serial_number=lens_sn);
+filter_ID = data['filter_ID']
+filter_ID_code = interactive(d, filter_ID_code=filter_ID);
+serial_into_USBhub = data['serial_into_USBhub_port']
+serial_into_USBhub_port = interactive(e, serial_into_USBhub_port=serial_into_USBhub);
+camera_into_serial = data['camera_into_serial_port']
+camera_into_serial_port = interactive(f, camera_into_serial_port=camera_into_serial);
+USBhub = data['USBhub_SN']
+USBhub_SN = interactive(j, USBhub_SN=USBhub);
+camera_into_USBhub = data['camera_into_USBhub_port']
+camera_into_USBhub_port = interactive(k, camera_into_USBhub_port=camera_into_USBhub);
+
+
+# In[9]:
+
+def h(add_device):
+    birger_SN = birger_serial_number.result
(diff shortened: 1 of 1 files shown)


wtgee:
Can you create an example notebook in `examples/notebooks/` that does as you describe?

moniquecockram:
There should already be one there, they were uploaded at the same time

wtgee:
> There should already be one there, they were uploaded at the same time

It's not part of this PR. Only one file.

I'll take a look at the rest of this in a bit.

moniquecockram:
The example notebook seems to be appearing now for me, let me know if it still isn't there.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
+
+# coding: utf-8
+
+# In[1]:
+
+from __future__ import print_function
+import yaml
+from ipywidgets import interact, interactive, fixed, interact_manual
+import ipywidgets as widgets
```
Physarah: Maybe swap around the format of your imports so all the 'import's are at the top and then 'from's?

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
...
+                              'mounts' : 'POCS/resources/mounts'
+             },
+    'cameras' : {'hdr_mode' : True,
+               'auto_detect' : False,
+                'devices' : [None]}}
+
+
+# In[7]:
+
+def a(birger_serial_number):
```
Physarah: Not really sure why the functions are defined as a,b,c etc. Is there a reason? Can be a bit confusing.

moniquecockram: they needed to be seperate functions so that the widgets could each have different labels. I just chose those letters as it was simple to copy and paste and change, but if it is confusing I can easily change them to something more logical

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
...
+camera_into_serial_port = interactive(f, camera_into_serial_port=camera_into_serial);
+USBhub = data['USBhub_SN']
+USBhub_SN = interactive(j, USBhub_SN=USBhub);
+camera_into_USBhub = data['camera_into_USBhub_port']
+camera_into_USBhub_port = interactive(k, camera_into_USBhub_port=camera_into_USBhub);
+
+
+# In[9]:
+
+def h(add_device):
```
Physarah: Also, you might want to add a little description of what the functions do, args and returns etc. You can find how Wilfred likes this to be set out [here](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

scripts/generating_yaml.py
wtgee: Definitely name to something more logical.

scripts/generating_yaml.py
moniquecockram: Good idea, will do

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
...
+    yaml.dump(data_list,objFile1, default_flow_style=False, indent=4)
+    objFile1.close()
+
+    strOutFile = os.path.join("/var/huntsman-pocs/conf_files/huntsman_archive", archive_filename)
+    objFile = open(strOutFile, "w")
+    yaml.dump(data_list,objFile, default_flow_style=False, indent=4)
+    objFile.close()
+
+    return create_file
+
```
Physarah: There is also quite a bit of white space, which you can fix by doing something I learned called pep8 linting. The easiest way I find is by running [autopep8](https://pypi.python.org/pypi/autopep8). You can just run 
```
pip install --upgrade autopep8
```
And then with a terminal running in your environment 
```
autopep8 --in-place --aggressive --aggressive <filename>
```

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
...
+
+# In[11]:
+
+archive_filename
+
+
+# In[12]:
+
+def g(create_file):
```
Physarah: What does create_file do? You don't seem to use it again though it's returned. Doesn't seem to be used in the function either? Or do I not understand whats happening :P

scripts/generating_yaml.py
Physarah: Similar query to 'create_device'. 'add_device' doesn't seem to have a reason to be returned?

scripts/generating_yaml.py
moniquecockram: I think it's a bit hard to see what's happening because you aren't seeing the jupyter widgets that come up when it is run. The widgets work in an odd way, the function is defining what to do once the widget is clicked. So the function doesn't have a specific arg value, it just returns what the widget outputs. I'll try and explain it a bit more in the code if I can, too.

I think the return line came about from the earlier model I was following for those functions, but I have just run it after removing the return and it appears to be working fine, which does make sense.

Physarah: Okidoke, yeah I'm not very familiar with how the widgets work

scripts/generating_yaml.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+# In[6]:
+
+data_list = {'name' : 'HuntsmanSSO',
+             'log_file' : 'huntsman',
+             'location' : {'name' : 'Siding Spring Observatory',
+                           'latitude' : -31.27, #degrees
+                           'longitude': 149.06, # Degrees
+                          'elevation': 1000.0, # Meters
```
Physarah: Just adding this for reference like we discussed, the data list might be able to be imported straight from the config file.

moniquecockram:
All of these issues should now be fixed.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+        Args:
+            camera_into_USBhub_port (str) : the port number of the USB Hub that the camera is plugged into as selected from the widget.
+
+        Returns:
+            The result of the widget; the chosen USB port number
+
+        """
+        return camera_into_USBhub_port
+
+    birger_sn = data['birger_SN']
```
lspitler: Maybe provide a 1-line comment before this block briefly stating what it is about to do.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+    camera_into_USBhub = data['camera_into_USBhub_port']
+    camera_into_USBhub_port = interactive(
+        camera_to_usb_widget, camera_into_USBhub_port=camera_into_USBhub)
+
+    def add_device_widget(add_device):
+        """Function to add the details selected using the drop-down menu widgets to the 'data_list' dictionary.
+        The function is called by the widget and is then run when the user clicks on the widget button.
+
+        Args:
+            add_device : on clicking the widget, a device set is saved to the dict.
```
lspitler: `add_device` is not used anywhere. Is it needed?

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+
+        Returns:
```
lspitler: Maybe say `Output` like you do above instead of `Returns` as nothing is being returned.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+            Appends the data_list dict with the information chosen from the device information widgets
+
+        """
+        birger_SN = birger_serial_number.result
```
lspitler: WHen you move the `add_device_widget` out of your `run` function, it likely won't be able to access these variables any more. You might need to pass each and every one into `add_device_widget` from the `run` function.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+        filter_ID = filter_ID_code.result
+        serial_to_USBhub_port = serial_into_USBhub_port.result
+        camera_to_serial_port = camera_into_serial_port.result
+        USB_hub_SN = USBhub_SN.result
+        camera_to_USBhub_port = camera_into_USBhub_port.result
+
+        lens_name_dict = {'3360000099': 'name1', '3360000063': 'name2',
+                          '3360000087': 'name3', '2850000067': 'name4',
+                          '3150000110': 'name5', '5370000054': 'name6'}
+        # need to put the lens names in the above dict when they are finalised
```
lspitler: Add a formal Issue on github and remove this note.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+                                      'image_stabalisataion': lens_image_stabalisation[lens_SN]},
+                             'USB_hub_serial_number': USB_hub_SN,
+                             'camera_into_serial_adaptor_port': camera_to_serial_port,
+                             'serial_adaptor_into_USBhub_port': serial_to_USBhub_port,
+                             'camera_into_USBhub_port': camera_to_USBhub_port
+                             }
+        if data_list['cameras']['devices'] == [None]:
+            data_list['cameras']['devices'] = [additional_device]
+        else:
+            data_list['cameras']['devices'].append(additional_device)
```
lspitler: OK, looks like the final result is a `data_list` dictionary. Maybe this should be returned so the next function can use it.

lspitler: Or maybe you need to define a `global` variable named `data_list` outside of all functions in this file, near the top of the file so all functions can access it.

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+    archive_filename = '{}_{}.{}'.format('huntsman', datetime_str, 'yaml')
+
+    def save_file_widget(create_file):
+        """This function writes the 'data_list' dictionary to a .yaml text file.
+        The function is called by the widget and is run when the user clicks on the widget button.
+
+        Args:
+            create_file : on clicking the widget, the dict is saved to a .yaml file
+
+        Returns:
```
lspitler: Returns -> Output (or result?)

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+    date_info = datetime.datetime.today()
+    datetime_str = date_info.strftime('%Y_%m_%d_%H_%M')
+    archive_filename = '{}_{}.{}'.format('huntsman', datetime_str, 'yaml')
+
+    def save_file_widget(create_file):
+        """This function writes the 'data_list' dictionary to a .yaml text file.
+        The function is called by the widget and is run when the user clicks on the widget button.
+
+        Args:
+            create_file : on clicking the widget, the dict is saved to a .yaml file
```
lspitler: Do you need `create_file`?

scripts/generating_yaml.py
```diff
@@ -0,0 +1,286 @@
...
+                    for the local config file to be used by POCS
+
+                    and
+
+                '/var/huntsman-pocs/conf_files/huntsman_archive/huntsman_YYYY_mm_dd_hh_MM.yaml'
+                    for the archive of all version of the config files, with the date it was created in the filename
+
+        """
+
+        strOutFile1 = os.path.join(
```
lspitler: Can you pass in the directory location?

scripts/generating_yaml.py
```diff
@@ -1,213 +1,286 @@
+def run():
+    """This function runs all the code to generate the .yaml config files for the Huntsman-POCS system.
+    It displays the Jupyter widgets which the user can interact with to write and save the config files.
 
-# coding: utf-8
+    Files are saved in two locations:
+        '/var/huntsman-pocs/conf_files/huntsman.yaml' for the local file.
+        '/var/huntsman-pocs/conf_files/huntsman_archive/huntsman_YYYY_mm_dd_hh_MM.yaml' for the archive/version control file.
```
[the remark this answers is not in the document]
moniquecockram: It should be fine, it creates the new directory if it doesn't exist

PR #50, opened 2018-02-28
Add weather files to huntsman-pocs

jermainegug:
Can check the progress of these files in https://github.com/AstroHuntsman/PEAS.

### DIFF
--- a/conf_files/weather_config.yaml
+++ b/conf_files/weather_config.yaml
@@ -1,0 +1,120 @@
+weather:
+    station: mongo
+    aat_metdata:
+      name: AAT metdata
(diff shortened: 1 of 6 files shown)


AnthonyHorton:
@wtgee is this stuff still needed? Or has most/all of it been merged upstream?

wtgee:
> @wtgee is this stuff still needed? Or has most/all of it been merged upstream?

The SQM stuff is still unique as are some of the other classes. We are not running the other weather checks (e.g. SkyMapper, AAT) so whatever code we have on there is mostly untested.

I think this is all still outstanding actually. :(

PR #55, opened 2018-07-13
Merge changes from May Huntsman commissioning run

AnthonyHorton:
A new Huntsman commissioning run (and a bunch of frenzied coding) is imminent, so I'd like to merge the changes from the previous run. There's a lot to untangle here, and I suspect we won't want to keep it all, but I'm opening the PR to get the discussion going.


wtgee (approved):
There's a number of files that I don't think we want in the repositories. We should probably clean them up locally at the site as well but let's definitely keep them out of here. Can you remove those and then one or two other quick comments. Thanks!

Issue #60, opened 2018-08-17
add musca shutter control


lspitler:
We're work in this repo because its very huntsman specific. WE can move to POCS later on if we want.

Here are the main tasks:

- [x] Create a new branch based upon this issue from the `develop` branch.
- [x] Look at files here: https://github.com/panoptes/POCS/tree/develop/pocs/dome
- [x] Look at this file as an example of a serial comms dome control: https://github.com/panoptes/POCS/blob/develop/pocs/dome/astrohaven.py 
- [x] Create musca.py and inherit bisque.py dome and start writing in Musca commands
- [x] fix shutter musca status #82 
- [x] Ask @wtgee how to link the bisque dome with the musca dome
- [x] Request a non-testing Pull Request (use a `WIP` flag in PR name) review for structure and programming and doc strings
- [x] ~~@AnthonyHorton to determine if we really need a super-fancy protocol simulator like: https://github.com/panoptes/POCS/blob/develop/pocs/dome/protocol_astrohaven_simulator.py~~
- [ ] Add a check voltage query
- [ ] Test it on hardware
- [ ] Ponder what to do if OSX drops the bluetooth connection & how to re-establish connection.
- [ ] Write unit tests when hardware and a human is present in the dome to ensure it works
- [ ] Final Pull Request
- [ ] consider how or if to integrate bisque into musca class
- [ ] create heartbeat via status - send open cmd - modify open so it actually sends command

lspitler:
OK, @wtgee I think the next step is to create an instance of the dome on the control computer and test out these functions on a non-rainy day. Would you agree?

I poked around POCS a bit and found a few places where a dome `status()` method could start being put throughout (e.g. https://github.com/panoptes/POCS/blob/41fd0484772bd30b1cf3ab5b8cb0329223177f28/pocs/core.py#L169 ) and have ideas where to put a sort of dome shutter daemon (e.g. https://github.com/panoptes/POCS/blob/41fd0484772bd30b1cf3ab5b8cb0329223177f28/pocs/state/machine.py#L213 ), but would need more guidance from you on how to make it happen.

lspitler:
It was decided that we don't need to use a complex simulator for tests - we'll try a hardware test sequence at the dome.

danjampro:
Complete.

PR #61, opened 2018-08-30
huntsman dome control WIP

bazkiaei:
Work for #60


wtgee (commented):
Seems fine for starting!

wtgee (approved):
I would like to work on #587 to clean up some of the status call items but that can come later. This is a good step to getting us going, thanks!

PR #65, opened 2018-09-17
backup working notebook from Sept 2018

huntsmanscope:
Would someone mind reviewing this quickly. Just want to ensure this notebook is somewhere, even if we're about to go away from this mode. Handy dark script too.


lspitler:
@wtgee I think you've already done this? So I might delete this PR?

wtgee:
I don't think we need this here. Let's close the PR but keep the branch. I'm going to get to some of the cleanup  from the September run today.

PR #70, opened 2018-09-27
Adding in the Hue Lighting Control

wtgee:
Some old code from @Physarah


lspitler:
@wtgee would you like a full review on this or has it already been screened?

wtgee:
> @wtgee would you like a full review on this or has it already been screened?

We need a full review, including by myself. I suspect we will need to do some work before merge. This was mostly to not lose it in the public switchover.

lspitler:
OK, we'll get to this eventually then.

AnthonyHorton:
One obvious thing: no tests. Probably not possible to have unit tests for this because it would require the hardware to be present, in which case the un-testable code should be excluded from the test coverage calculation.

AnthonyHorton:
This is failing tests right now, but that seems to be due to an unrelated bug in the weather system or, more likely, the weather system tests.

PR #72, opened 2018-09-28
Fix python version for travis

wtgee:


lspitler (approved):

wtgee:
As a general policy you should wait for tests to complete even if you are sure they will pass.

Because sometimes they don't. :)

lspitler:
@wtgee Ha! Should I re-open the branch or do we need another one don't we?

wtgee:
We'll have to do a new one, which I forgot about. Thanks for reminder! ;)

Issue #69, opened 2018-09-27
Autofocus improvements


AnthonyHorton:
Experience during Huntsman commissioning shows that while the autofocus routine is working well there are still a few improvements that would help to make it more robust, paving the way for remote observing.

- [x] Automatically run a coarse focus prior to the first observations of the night, then run a fine focus prior to each block of observations (latter already happens).
- [ ] If best focus appears to be outside the range of the focus sweep automatically extend the range with additional focus images and try again.
- [ ] Determine whether there are stars in the image and if not (e.g. vignetted by dome, clouds) discard the results and issue a warning.
- [ ] Report whether focus failed (e.g. no stars) and, if successful, some measure of confidence in the results.
- [ ] Add information about most recent successful focus (time, initial, final, confidence) to FITS headers
- [ ] Add an option to follow autofocus by deliberately defocusing by a set amount, for bright star photometry.
- [ ] Focus prediction based on ambient temperature and empirical temperature - focus relation (longer term)
- [ ] consider discussion here: https://aao-org.slack.com/archives/GQ9F8Q5FY/p1602583198298800
- [ ] change focus setting based upon current filter https://aao-org.slack.com/archives/GKNFB1TEH/p1611806150375900

lspitler:
![](https://github.trello.services/images/mini-trello-icon.png) [address focuser issues](https://trello.com/c/BQy6WV3F/103-address-focuser-issues)

AnthonyHorton:
Implementing the first item with https://github.com/panoptes/POCS/pull/659 and #75

lspitler:
See also:

https://github.com/AstroHuntsman/huntsman-pocs/issues/202

https://github.com/AstroHuntsman/huntsman-pocs/issues/199

lspitler:
memory issues https://github.com/panoptes/POCS/issues/1028

lspitler:
See also https://github.com/panoptes/POCS/pull/1029

PR #75, opened 2018-10-02
Add coarse focus to evening calibrations

AnthonyHorton:
Adds a coarse focus off all cameras to the evening calibration state, after flat fielding and nautical twilight, in order to implement the first item in #69.

Also took the opportunity to refactor the waiting for a specific Sun altitude into a utility function, and edited a few message strings ('sunset' -> 'twilight').


wtgee (changes requested):
Hmm, a lot of the same thing is going on over here: https://github.com/panoptes/POCS/pull/541/files#diff-a318b148222fe28a50df3aa2824429fdR522. It's a little more convoluted over there as it is pulled into the pocs machinery itself and reads from the config.

It keeps the `pocs` objects out of utils file, which I think we should enforce. The `pocs` items are only used in state files as it is.

Can we just add the coarse focus after the flat-fielding for this PR and then we'll add in the appropriate wait after?

AnthonyHorton:
Ah, wasn't aware of the overlap with a WIP in POCS itself. Was just keen to make sure it was dark enough for the coarse focus to work, and felt I shouldn't just add a 3rd nearly identical `while` loop to `calibrating.py`. I would have just passed the `observatory` or `observer` to my utility function, but then I wouldn't be able to keep the `pocs.say`.

AnthonyHorton:
@wtgee Compromise temporary solution, move the `wait_for_sun_alt` utility function into `huntsman-pocs/states/calibrating.py` instead of `huntsman-pocs/utils/flats.py`?   Keeps the `pocs` object out of `utils` and avoids any overlap with the POCS WIP while providing the desired functionality in the meantime.  Once the WIP gets merged into POCS we can chuck this out of the `huntsman-pocs` state machine and use the new functions.  Going to want to do a bunch of that sort of thing anyway to stop `huntsman-pocs` duplicating functionality that's been added to POCS.

AnthonyHorton:
Doing some tinkering to get Travis and coverage working. Will spin out into a separate PR.

states/huntsman/calibrating.py
```diff
@@ -3,6 +3,42 @@
...
+        min_altitude (astropy.units.Quantity, optional):
+        max_altitude (astropy.units.Quantity, optional):
+        message (str):
+    """
+    if min_altitude is None and max_altitude is None:
+        raise ValueError("At least one of min_altitude & max_altitude must be given")
+    if min_altitude is None:
+        min_altitude = -90
+    if max_altitude is None:
+        max_altitude is +90
```
wtgee: `max_altitude = 90`

wtgee (approved):
LGTM although I see tests are failing.

states/huntsman/calibrating.py
AnthonyHorton: OK.

AnthonyHorton:
Passing now, thanks to #79!

PR #79, opened 2018-10-02
Fixing tests

### LINKED ISSUE #77: Fix test_huntsman tests
The `huntsman-pocs` test suite does not run successfully because the tests in `huntsman/tests/test_huntsman.py` have not been updated to keep track of changes in `panoptes/POCS`.  Currently the tests hang indefinitely in `test_run_wait_until_safe`.

The tests here are largely a duplicate of those in `panoptes/tests/test_pocs.py`, and they should be apart from a few parameter values. Several of the tests in `huntsman/tests/test_huntsman.py` have now diverged significantly from the `panoptes/POCS` versions, though. The troublesome `test_run_wait_until_safe` is one example. 

I've been attempting to get the test suite running in Travis CI again in #76 but it's clear that a number of tests need fixing, not just config files, so I'm opening a separate Issue for fixing tests.

wtgee:
This is a mish-mash of fixes to get testing working. Includes a better (hopefully) way of loading config files. Ideally would come up with a more generic way for derivatives of POCS to do this.

Note that I've forced the tests with the mongo db only, unlike POCS, which does all three db types

Fixes #77 (still could use plenty more tests).

### DIFF
--- a/huntsman/observatory.py
+++ b/huntsman/observatory.py
@@ -11,12 +11,12 @@ from pocs.scheduler import constraint
 from pocs.scheduler.observation import Field
 from pocs.utils import error
 from pocs import utils
-from pocs.utils import config as cfg
 from pocs.utils.images import fits as fits_utils
 
 from huntsman.guide.bisque import Guide
 from huntsman.scheduler.observation import DitheredObservation
 from huntsman.utils import dither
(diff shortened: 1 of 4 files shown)


AnthonyHorton (approved):
Great! LGTM.

PR #80, opened 2018-10-05
Add support for distributed Cameras

AnthonyHorton:
_Moved_ _from_ _https://github.com/panoptes/POCS/pull/547_

Addressing Issue https://github.com/panoptes/POCS/issues/546 by adding scripts and wrappers to enable distributed control of camera arrays using the Pyro4 library.

- [X] Pyro name server startup
- [X] Pyro camera server startup
- [X] Automatic discovery of remote cameras
- [X] Initialise remote cameras, get UIDs
- [X] Initiate exposures on remote cameras
- [X] Transfer files from remote cameras after exposures
- [x] Autofocus of remote cameras
- [x] `take_observation` using remote cameras
- [x] Access remote camera properties
- [x] Set remote camera properties


conf_files/pyro_camera.yaml
```diff
@@ -0,0 +1,99 @@
+# Config file for a distributed camera & focuser.
+#
+# Contains a default configuration with a simulated camera and focuser for testing
+# and example configurations for other camera and focusers (commented out).
+#
+# Do not edit this file, instead copy one of the examples to pyro_camera_local.yaml
```
wtgee: I almost want this at the top and in all caps. And it will probably still happen. :)

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+
+from pocs.camera.camera import AbstractCamera  # pragma: no flakes
+from pocs.camera.camera import AbstractGPhotoCamera  # pragma: no flakes
+
+from huntsman.camera.pyro import Camera as PyroCamera
+
+from pocs.utils import logger as logger_module
+
+
+def list_connected_cameras():
```
wtgee: This isn't different, right? I realize you were probably just copying straying over, but let's just remove from here as the other is still available to us via POCS.

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+def list_distributed_cameras(ns_host=None, logger=None):
+    """Detect distributed cameras.
+
+    Looks for a Pyro name server and queries it for the list of registered cameras.
+
+    Args:
+        host (str, optional): hostname or IP address of the name server host. If not given
+            will attempt to locate the name server via UDP network broadcast.
+        logger (logging.Logger, optional): logger to use for messages, if not given will
+            ise the root logger.
```
wtgee: s/ise/use/

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Returns:
+        dict: Dictionary of distributed camera name, URI pairs
```
wtgee: maybe the word "detected" in here?

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+                    cam.is_primary = True
+                    primary_camera = cam
+                    is_primary = ' [Primary]'
+
+                logger.debug("Camera created: {} {}{}".format(
+                    cam.name, cam.uid, is_primary))
+
+                cameras[cam_name] = cam
+
+    # Distributed camera creation
```
wtgee: As far as I can tell lines 220-228 are the only lines different from what would be in POCS. Would it work to move these lines into `create_distributed_cameras` (maybe with rename to `create_distributed_cameras_from_config`) and then call the the functions sequentially? Would have to deal with the primary camera preference in some fashion but you already assume here that the distributed should override the local.

Would just be nice to reduce all this to only what we need separately.

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+    logger.debug("Primary camera: {}", primary_camera)
+    logger.debug("{} cameras created", len(cameras))
+
+    return cameras
+
+
+def create_distributed_cameras(camera_info, logger=None):
+    """Create distributed camera object(s) based on detected cameras and config
+
+    Creates a pocs.camera.pyro.Camera object for each distributed camera detected.
```
wtgee: Since I've been doing a lot of the docs stuff lately, I would add the backticks around the `pocs...Camera` here.

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Args:
+        camera_info: 'cameras' section from POCS config
+        logger (logging.Logger, optional): logger to use for messages, if not given will
+            use the root logger.
+
+    Returns:
+        OrderedDict: An ordered dictionary of created camera objects, with the
```
wtgee, line 143: This doesn't match what is returned.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+from threading import Event
+from threading import Timer
+from threading import Thread
+from threading import Lock
+import subprocess
+
+from astropy import units as u
+import Pyro4
+
+from pocs.base import PanBase
```
wtgee: I'm not seeing this anywhere?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+        super().__init__(name=name, port=uri, model=model, *args, **kwargs)
+
+        # Connect to camera
+        self.connect(uri)
+
+# Properties
+
+    @AbstractCamera.uid.getter
+    def uid(self):
+        # Neet to overide this because the base class only returns the 1st 6 characters of the
```
wtgee: s/Neet/Need/

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+    """
+    def __init__(self,
+                 uri,
+                 name='Pyro Camera',
+                 model='pyro',
+                 *args, **kwargs):
+        super().__init__(name=name, port=uri, model=model, *args, **kwargs)
+
+        # Connect to camera
+        self.connect(uri)
```
wtgee: You don't want to store `uri` so can can call `connect()` below without uri again? (Or a thin-wrapper `reconnect` in addition?)

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+        try:
+            self._proxy = Pyro4.Proxy(uri)
+        except Pyro4.errors.NamingError as err:
+            msg = "Couldn't get proxy to camera {}: {}".format(self.name, err)
+            warn(msg)
+            self.logger.error(msg)
+            return
+
+        # Set sync mode
+        Pyro4.async(self._proxy, async=False)
```
wtgee: Would we ever want async?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+
+        if isinstance(timeout, u.Quantity):
+            timeout = timeout.to(u.second)
+            timeout = timeout.value
+        if timeout is not None:
+            timeout = float(timeout)
+
+        dir_name, base_name = os.path.split(filename)
+        # Make sure dir_name has one and only one trailing slash, otherwise rsync may fail
+        dir_name = dir_name.rstrip('/') + '/'
```
wtgee: Would `os.path.normpath` do this?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+        self.logger.debug('Taking {} second exposure on {}: {}'.format(
+            seconds, self.name, base_name))
+        # Remote method call to start the exposure
+        exposure_result = self._proxy.take_exposure(seconds=seconds,
+                                                    base_name=base_name,
+                                                    dark=bool(dark),
+                                                    *args,
+                                                    **kwargs)
+        # Tag the file transfer on the end.
+        exposure_result = exposure_result.then(self._file_transfer, dir_name)
```
wtgee, line 186: I still think I would prefer to go the extra setup mile with ssh keys and then just map the directories so that the file transfer is transparent. The rsync works for now and it looks like this line make it easy to switch out (or remove) in the future.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+            threading.Event: Event that will be set when exposure is complete
+
+        """
+        assert self.is_connected, self.logger.error("Camera must be connected for take_exposure!")
+
+        assert filename is not None, self.logger.warning("Must pass filename for take_exposure")
+
+        # Want exposure time as a builtin type for Pyro serialisation
+        if isinstance(seconds, u.Quantity):
+            seconds = seconds.to(u.second)
```
wtgee: Can these two lines be compressed?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+            seconds = seconds.value
+        seconds = float(seconds)
+
+        if isinstance(timeout, u.Quantity):
+            timeout = timeout.to(u.second)
```
wtgee: Same.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+                            'take_dark': take_dark,
+                            'thumbnail_size': thumbnail_size,
+                            'merit_function': merit_function,
+                            'merit_function_kwargs': merit_function_kwargs,
+                            'mask_dilations': mask_dilations,
+                            'coarse': coarse,
+                            'make_plots': make_plots}
+        autofocus_kwargs.update(kwargs)
+
+        focus_dir = os.path.join(os.path.abspath(self.config['directories']['images']), 'focus/')
```
wtgee, line 306: Is everything above just for serialization?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+    def _clean_directories(self, source):
+        """
+        Clean up empty directories left behind by rsysc.
+        """
+        user_at_host, path = source.split(':')
+        path_root = path.split('/./')[0]
+        try:
+            result = subprocess.run(['ssh',
+                                     user_at_host,
+                                     'find {} -empty -delete'.format(path_root)],
```
wtgee, line 347: Wait, what? You can delete empty directories like that via `find`? I always thought you would have to send to `xargs` or something.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+
+@Pyro4.expose
+@Pyro4.behavior(instance_mode="single")
+class CameraServer(object):
+    """
+    Wrapper for the camera class for use as a Pyro camera server
+    """
+    def __init__(self):
+        # Pyro classes ideally have no arguments for the constructor. Do it all from config file.
+        self.config = load_config(config_files=['pyro_camera.yaml'])
```
wtgee: Probably should make the file a parameter just in case.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+        self.name = self.config.get('name')
+        self.host = self.config.get('host')
+        self.port = self.config.get('port')
+        self.user = os.getenv('PANUSER', 'panoptes')
```
wtgee: `huntsman` as user here.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+        return temperature.to(u.Celsius).value
+
+    @property
+    def ccd_set_point(self):
+        temperature = self._camera.ccd_set_point
+        return temperature.to(u.Celsius).value
+
+    @ccd_set_point.setter
+    def ccd_set_point(self, set_point):
+        self._camera.ccd_set_point = set_point
```
wtgee, line 453: Looks like you need to insure this is a `Quantity` or add.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,502 @@
...
+
+    @property
+    def ccd_cooling_enabled(self):
+        return self._camera.ccd_cooling_enabled
+
+    @ccd_cooling_enabled.setter
+    def ccd_cooling_enabled(self, enabled):
+        self._camera.ccd_cooling_enabled = enabled
```
wtgee, line 461: Probably force `bool` here too.

huntsman/tests/conftest.py
```diff
@@ -52,3 +57,81 @@ def db():
 @pytest.fixture
 def data_dir():
     return '{}/pocs/tests/data'.format(os.getenv('POCS'))
+
+
+def end_process(proc):
```
wtgee, line 62: Maybe a docstring explaining this deeply nest kill structure. :)

huntsman/tests/test_camera.py
```diff
@@ -0,0 +1,347 @@
...
+    camera.take_exposure(filename=fits_path, dark=True, blocking=True)
+    assert os.path.exists(fits_path)
+    # If can retrieve some header data there's a good chance it's a valid FITS file
+    header = fits_utils.getheader(fits_path)
+    assert header['EXPTIME'] == 1.0
+    assert header['IMAGETYP'] == 'Dark Frame'
+
+
+@pytest.mark.filterwarnings('ignore:Attempt to start exposure')
+def test_exposure_collision(camera, tmpdir):
```
wtgee, line 170: Still needed here?

huntsman/tests/test_observatory.py
```diff
@@ -0,0 +1,380 @@
...
+    Tests that require real hardware should be marked with the appropriate
+    fixture (see `conftest.py`)
+    """
+    return hardware.get_all_names(without=['night'])
+
+
+@pytest.fixture
+def observatory(config, simulator):
+    """Return a valid Observatory instance with a specific config."""
+    cameras = create_cameras_from_config(config)
```
wtgee: Looks like this was supposed to change?

huntsman/utils/pyro/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+            from the name server if they cannot be connected. If not given no autocleaning will
+            be done.
+    """
+    try:
+        # Check that there isn't a name server already running
+        name_server = Pyro4.locateNS()
+    except errors.NamingError:
+        if not host:
+            # Not given an hostname or IP address. Will attempt to work it out.
+            host = get_own_ip(verbose=True)
```
wtgee, line 100: Can we just run on the localhost if none is given? Or is auto-detection the assumed usage?

huntsman/utils/pyro/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            host = str(host)
+
+        if port:
+            port = int(port)
+        else:
```
wtgee: `else` seems unnecessary.

huntsman/utils/pyro/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+
+def run_camera_server(ignore_local):
+    """
+    Runs a Pyro camera server.
+
+    The camera server should be run on the camera control computers of distributed cameras. The
+    camera servers should be started after the name server, but before POCS.
+
+    Args:
+        ignore_local (bool, optional): If True use the default $POCS/conf_files/pyro_camera.yaml
```
wtgee: s/$POCS/$HUNTSMAN_POCS/

huntsman/utils/pyro/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+        uri = daemon.register(CameraServer)
+        print('Camera server registered with daemon as {}'.format(uri))
+        name_server.register(config['name'], uri, metadata={"POCS",
+                                                            "Camera",
+                                                            config['camera']['model']})
+        print('Registered with name server as {}'.format(config['name']))
+        print('Starting request loop... (Control-C/Command-C to exit)')
+        try:
+            daemon.requestLoop()
+        finally:
```
wtgee, line 151: Do you need to catch the `KeyboardInterrupt`?

scripts/pyro_name_server.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+import argparse
+
+from huntsman.utils.pyro import run_name_server
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--host", help="hostname or IP address to bind the server on")
+    parser.add_argument("--port", help="port number to bind the server on")
+    parser.add_argument("--autoclean", help="interval between registration autoclean (0=disabled)",
```
wtgee, line 16: Did we want to enable this by default or do you think not needed?

wtgee (commented):
Whew! That was a long review which means that was even longer for you! 

Thanks for getting this moved over. Submitting just as comment for now but can approve if you would rather merge before vacation. 

Seems like there is a bit we could do to eliminate code duplication between here and POCS, but I'm happy to do that after the fact.

conf_files/pyro_camera.yaml
AnthonyHorton: A compromise, I've made it all caps but didn't move it :-)

huntsman/camera/__init__.py
AnthonyHorton: Yes, this entire file is just copied over verbatim from the branch at `panoptes/POCS`. A fair bit of unnecessary code duplication that I hadn't made any attempt to reduce yet. Replacing this with an import now.

huntsman/camera/__init__.py
AnthonyHorton: Well, these few lines plus a small change about how the default primary camera gets chosen (I can't assume that there is a camera named `Cam00` so I sort the camera names and choose the first one). But otherwise it is identical to the POCS version.

Yes, it was the awkwardness of keeping track of whether a primary camera has been selected yet that stopped me from making these two entirely separate functions in the first place. I'll have a think about how to work around that.

It doesn't really override a local primary camera with a distributed one, as there will be at most one primary camera specified in the config file so the distributed camera creation will only set the primary camera is the local camera creation hasn't.

huntsman/camera/__init__.py
AnthonyHorton: Good point.

huntsman/camera/__init__.py
AnthonyHorton: Yep.

huntsman/camera/__init__.py
AnthonyHorton: Yes.

huntsman/camera/__init__.py
AnthonyHorton: Oops, forgot to add the primary camera return value.

huntsman/camera/pyro.py
AnthonyHorton: Good catch. Must have been using it at some point, but not any more.

huntsman/camera/pyro.py
AnthonyHorton: There's probably a lot of instances of 'Neet' scattered through the Camera classes. Most of them have this same docstring...

huntsman/camera/pyro.py
AnthonyHorton: Probably a good idea, if for no other reason than to order to allow us to later on to defer the `connect()` instead of calling it from within `__init__()`.   That's something we might want to do, because it's the `connect()` that triggers the remote object creation and, for real cameras, the lengthy hardware initialisation process.

huntsman/camera/pyro.py
AnthonyHorton: We do, for the `take_exposure` and `autofocus` methods. I could just use `Thread`s for those, but instead used Pyro's built in asynchronous results functionality.

huntsman/camera/pyro.py
AnthonyHorton: If I'm reading the docs right `os.path.normpath` would replace the `rstrip` (as well as doing some extra tidying if needed) but I still need the `+ '/'`

huntsman/camera/pyro.py
AnthonyHorton: Yep, would be easy to change. ssh keys are already set up, otherwise rsync would be asking for passwords here.

huntsman/camera/pyro.py
AnthonyHorton: Looking at is again I see that I set synchronous mode in `connect()`, and set asynchronous mode at the start of both `take_exposure` and `autofocus`, but I don't switch it back to synchronous at the end of those methods. That ought to break the property access methods if called after an exposure. Hmmm.

AnthonyHorton: Oh, there's a change in naming here too, because `async` is a reserved keyword in Python 3.7. I've changed to the new names.

huntsman/camera/pyro.py
AnthonyHorton: Yes. Yes they can.

huntsman/camera/pyro.py
AnthonyHorton: Ditto.

huntsman/camera/pyro.py
AnthonyHorton: Yes, that's me doing my best to ensure either successful serialisation with Pyro's default serialiser, or an informative error. I added most of this today in response to the `np.int64` fiasco!

huntsman/camera/pyro.py
AnthonyHorton: Neat trick, huh? I found this is an answer on StackOverflow, of course.

huntsman/camera/pyro.py
AnthonyHorton: With Pyro the recommended approach is to pass the class to the daemon rather than an object, and let the daemon handle object creation. That precludes passing any arguments to it, so I didn't even bother putting any in the constructor here. It is possible to pass an object instead though, which would enable overriding the config file name.

huntsman/camera/pyro.py
AnthonyHorton: Yep.

huntsman/camera/pyro.py
AnthonyHorton: The setter in the underlying `_camera` does that.

huntsman/camera/pyro.py
AnthonyHorton: Same as above.

huntsman/tests/conftest.py
AnthonyHorton: Did have a few problems with these processes being left hanging after running tests....

huntsman/tests/test_camera.py
AnthonyHorton: Hmmm, These Pyro cameras don't emit the warning that the other Camera classes do (shame!), so I don't need the warnings filter. The test is still useful though, just to make sure that trying to start another exposure while one is in progress doesn't break anything.

I just leave it up to Pyro to handle any 'exposure collisions'. I've configured the camera servers to create only a single camera object, to handle remote method calls sequentially ('multiplex' mode), and made the server side `take_exposure` method blocking. If you try to start another exposure while one is in progress Pyro will just put it in a queue until the remote camera is ready.

huntsman/tests/test_observatory.py
AnthonyHorton: Yes. This entire file is just copied verbatim from the POCS version right now. There will be a few things that needs changing, and other buts that are redundant with `test_huntsman.py`. Haven't had chance to go through it yet.

huntsman/utils/pyro/__init__.py
AnthonyHorton: Running on localhost is 'special' for Pyro servers (including the name server), it disable the broadcast responder so they can't be reached from other machines on the network. Good for testing with simulators, but not a very useful default otherwise.

I see the auto detection as a convenience, really. It's probably not 100% reliable for complicated networking configurations where the machine has multiple IP address. For operational systems with fixed IPs it would be better to set the name server host manually via config file.

huntsman/utils/pyro/__init__.py
AnthonyHorton: It does. The only effect here is to convert a value of `port` that evaluates to `False` (e.g. `0`) into `None`.  I'm not sure why I would want that to happen, though.

huntsman/utils/pyro/__init__.py
AnthonyHorton: Yes.

huntsman/utils/pyro/__init__.py
AnthonyHorton: I tried doing that, but it didn't work as I expected. The `except` block didn't get executed. So I just put all the cleanup in `finally` and that did work.

scripts/pyro_name_server.py
AnthonyHorton: I think it should be enabled by default.  With the `finally` in the camera server they will now de-register themselves from the name server if you shut them down, but that doesn't handle more messy situations (network problems, Pi power failure, out of memory errors, etc.).  The name server autoclean is a handy catch-all backup.

AnthonyHorton:
Thanks for the detailed and quick review! I don't want to push through the merge just yet, though. At the very least want to get tests passing, and ideally remove the overlap and inconsistencies between `pocs.tests.test_observatory.py`, `huntsman.tests.test_observatory.py` & `huntsman.tests.test_huntsman.py`.

AnthonyHorton:
Looking at it a bit more I don't think I actually need anything from `test_observatory.py`, apart from `test_pyro_camera`.  So I've moved that into `test_huntsman.py` and removed `test_observatory.py`. Bingo! Passing tests and a diff that's ~400 lines smaller!

huntsman/camera/__init__.py
AnthonyHorton: OK, reduced code duplication a fair bit by importing `pocs.camera.create_cameras_from_config` to handle the local camera creation.

huntsman/camera/__init__.py
AnthonyHorton: Now moot because I took the primary camera part out again.

AnthonyHorton:
OK, reduced duplication of code from POCS in `huntsman/camera/__init__.py` and knocked another ~100 lines off the diff. Can I get a re-review, please, @wtgee? I think this might be OK to merge now.

Haven't done anything for those three unchecked items in the the PR description yet, but I think it's OK for them to be separate Issues & PRs.

huntsman/camera/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+
+    # Find primary camera
+    primary_camera = None
+    for camera in cameras.values():
+        if camera.is_primary:
+            primary_camera = camera
+
+    # If no camera was specified as primary use the first
+    if primary_camera is None:
+        camera_names = sorted(cameras.keys())
```
wtgee, line 122: Minor point, but since it's an OrderedDict do you want to skip the `sorted` and use the first one placed in the dict? I think it comes out the same anyway.

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,498 @@
...
+        Take exposure for a given number of seconds and saves to provided filename.
+
+        Args:
+            seconds (u.second, optional): Length of exposure
+            filename (str, optional): Image is saved to this filename
+            dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
+            blocking (bool, optional): If False (default) returns immediately after starting
+                the exposure, if True will block until it completes.
+            timeout (u.second, optional): Length of time beyond the length the exposure to wait
+                for exposures to complete. If not given will wait indefinitely.
```
wtgee, line 149: I'm still on the fence about *ever* allowing indefinite waits but it's fine for now. Maybe we need an issue for it.

huntsman/camera/pyro.py
wtgee: I assume the trailing `/` is for `rsync`?

huntsman/camera/pyro.py
```diff
@@ -0,0 +1,498 @@
...
+
+        return autofocus_event
+
+# Private Methods
+
+    def _clean_directories(self, source):
+        """
+        Clean up empty directories left behind by rsysc.
+        """
+        user_at_host, path = source.split(':')
```
wtgee, line 342: There's some weirdness with the path going on. Maybe an example of what the `source` input is here.

huntsman/tests/test_camera.py
```diff
@@ -0,0 +1,347 @@
+import pytest
```
wtgee, line 4: Do you want to just name this `test_pyro_camera.py` and then we can leave it for only Pyro items (if we ever add other camera types just to Huntsman)?

huntsman/tests/test_camera.py
wtgee: Got it. i was thinking this was testing the collision of multiple cameras (a la the SBIG simultaneous readout issue)

huntsman/tests/test_huntsman.py
```diff
@@ -225,6 +225,22 @@ def test_is_weather_safe_no_simulator(pocs, db):
     assert pocs.is_weather_safe() is False
 
 
+def test_pyro_camera(config, camera_server):
+    conf = config.copy()
+    conf['cameras'] = {'distributed_cameras': True}
+    simulator = hardware.get_all_names(without=['camera'])
+    conf['simulator'] = simulator
+    cameras = create_cameras_from_config(conf)
+    obs = Observatory(cameras=cameras,
```
wtgee, line 234: Note https://github.com/panoptes/POCS/pull/679 and the comments.

huntsman/utils/pyro/__init__.py
wtgee: I don't see this change.

huntsman/utils/pyro/__init__.py
wtgee: Maybe:

```python
try:
    daemon...
except KeyboardInterrupt:
    logger.info('Stopping  name server')
finally:
    # Cleanup
```

Not necessary though.

wtgee: That would only be so that we could add an *additional* `except Exception` so we could capture when it fails for some other reason. But again, not necessary for this PR.

scripts/pyro_camera_server.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python
+"""
+Script to run a Pyro camera server. This should be run on the camera control computer for a
+distributed camera. The configuration for the camera is read from the pyro_camera.yaml/
+pyro_camera_local.yaml config file. The camera servers should be started after the name server,
+but before POCS.
```
wtgee, line 6: Could we have this script check for a running name server and refuse to start if so? (Can be separate PR)

wtgee (approved):
Great! Looks good. Thanks for a ton of work to get this done in the last few days. Have a good vacation!

huntsman/camera/__init__.py
AnthonyHorton: I guess it depends on whether we want to default to the first camera added (which will be `Cam00` if there are any local cameras) or to the first camera alphabetically...

huntsman/camera/pyro.py
AnthonyHorton: At the least we should be enabling these timeouts in the testing and in the default config files to encourage their use. Will revisit that as part of the general issue of timeouts for all the cameras' threaded actions.

huntsman/camera/pyro.py
AnthonyHorton: Ah, yes, the directory cleanup follows the file transfer in the call chain, which just passes on the same `source` argument that the file transfer uses, which is in `user@host:/directory/subdirectory/filename` format ready for rysnc.

huntsman/tests/test_camera.py
AnthonyHorton: Inclined to leave as is to emphasise that this is a lightly edited copy of `pocs/tetsts/test_camera.py`, and is supposed to be kept in sync with that. In fact I'll add a comment about that.

huntsman/camera/pyro.py
AnthonyHorton: Yes, that's right.

huntsman/utils/pyro/__init__.py
AnthonyHorton: Fixed another instance of the same thing and missed this one!

huntsman/tests/test_huntsman.py
AnthonyHorton: I'll have to return to that one. Not sure I understand the underlying Iissue fully.

scripts/pyro_camera_server.py
AnthonyHorton: The `run_camera_server` function that the script calls does do exactly that. Might be neater to have the script do it instead, though.

PR #82, opened 2018-10-09
created and now use shutter status property

lspitler:
@wtgee take a quick look and let me know if this is the vision you had for a generic status property fo the domes.

### DIFF
--- a/huntsman/dome/musca.py
+++ b/huntsman/dome/musca.py
@@ -2,6 +2,7 @@
 
 import time
 
+from pocs.utils import CountdownTimer
 from pocs.dome.bisque import Dome as BisqueDome
 from pocs.dome.abstract_serial_dome import AbstractSerialDome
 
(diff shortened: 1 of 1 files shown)


huntsman/dome/musca.py
```diff
@@ -63,24 +63,33 @@ def __init__(self, *args, **kwargs):
...
         return v == Protocol.DOOR_CLOSED
 
+    @property
+    def status(self):
+        """A dictionary containing all status info for dome.
+
+        TODO: Add other info (e.g. dome moving)
+        TODO: Auto-update this every ~60s
+        """
+        return(self._get_shutter_status_dict())
```
wtgee: I think what we should do is set up a private variable `_status` as well as a timer, e.g. something like

```python
from pocs.utils import CoutdownTimer

...

def __init__(self):
    # other stuff
    self._status = dict()
    self._status_delay = 5 # seconds
    self._status_timer = CountdownTimer(self._status_delay)

def status(self):
    if self._status_timer.expired():
        self._status = self._get_shutter_status_dict()
        self._status_timer = CountdownTimer(self._status_delay)

    return self._status
```

huntsman/dome/musca.py
```diff
@@ -63,24 +63,33 @@ def __init__(self, *args, **kwargs):
 
     @property
     def is_open(self):
-        v = self._get_shutter_status_dict()[Protocol.SHUTTER]
+        v = self.status[Protocol.SHUTTER]
```
wtgee: `status` is a method, so these will all be, e.g. `v = self.status()[Protocol.SHUTTER]`, which is a bit annoying. This is why we have gone back and forth on it being a property versus a method.

wtgee (approved):
Comments should be addressed before merging but approving.

lspitler:
Thanks @wtgee once the tests pass I'll merge.

PR #85, opened 2018-10-18
Update current_exp_num property name

wtgee:
Changing property name to follow POCS

Note: the flat fielding was resetting the exposure number manually, which is no longer happening. Might see some odd things with flat field names after this.


lspitler (approved):
Looks good. Let's watch out for the counting issue...

PR #87, opened 2018-10-18
Properly reread targets file

wtgee:
The property is now read from the config, which we set to True by default.

### DIFF
--- a/conf_files/huntsman.yaml
+++ b/conf_files/huntsman.yaml
@@ -16,6 +16,7 @@ location:
 scheduler:
     type: dispatch
     fields_file: targets.yaml
+    check_file: True
 directories:
     base: /var/huntsman
     data: data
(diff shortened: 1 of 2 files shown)


conf_files/huntsman.yaml
```diff
@@ -16,6 +16,7 @@ location:
 scheduler:
     type: dispatch
     fields_file: targets.yaml
+    check_file: True
```
lspitler, line 19: Rename to make it clearer what it does. I guess this suggestion might have flow-on variables elsewhere in the code?
```suggestion
    reload_file_before_schedule: True
```

lspitler (approved):
one suggestion... fine otherwise.

conf_files/huntsman.yaml
wtgee: This comes down from the POCS side of things so we would need to make this change upstream.

PR #88, opened 2018-10-18
Fixing pyro result

wtgee:
This is the fix that was implemented during a testing run. This isn't really a fix for the problem but at least lets us work. Seems like Pyro was trying to serialize an astropy Quantity but it was really unclear where that was coming from. Ultimately the `_async_wait` just wants a success boolean so here we capture the error and then set to True. Could have bad side effects.

We'll need to merge this in to continue work but @AnthonyHorton might have some suggestions as to real fix.


lspitler (approved):
@wtgee will approve so you can merge (though check the checks - the coverage flags are not happy).

wtgee:
> @wtgee will approve so you can merge (though check the checks - the coverage flags are not happy).

I'm inclined to not worry about the coverage because this is not a real fix but a work around until we can figure out what is really going on with the serialization.

Issue #89, opened 2018-10-18
Dome rotation status


wtgee:
We need to know the dome rotation status so we don't start observing while dome is still rotating.

PR #90, opened 2018-10-18
Wait for autofocus events

### LINKED ISSUE #84: slew not blocking focus run at beginning of POCS.run()

wtgee:
Wait for events to complete properly.  Note that #89 is still a related problem.

Closes #84

### DIFF
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -96,7 +96,10 @@ def on_enter(event_data):
                      delay=60*3)
     try:
         pocs.say("Coarse focusing all cameras before starting observing for the night")
-        pocs.observatory.autofocus_cameras(coarse=True)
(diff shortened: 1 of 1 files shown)


lspitler (approved):
@wtgee 600s seems like a lot especially if this is happening between twilights & flats - will there be enough time for flats?

wtgee:
> @wtgee 600s seems like a lot especially if this is happening between twilights & flats - will there be enough time for flats?

Not sure i understand. This will start that autofocus routine and will terminate it after 10 minutes if it hasn't completed. I'm assuming that is enough time although my inline comment was about how maybe that is not enough. @AnthonyHorton might have a better sense if we would ever need more time for that.

But this happens after the flat fielding has already occurred and while waiting for astronomical twilight.

lspitler:
@wtgee all good then. I wasn't sure if it was before or after flats.

PR #97, opened 2018-11-24
Testing fixes

wtgee:
* Adding mpl for testing images (not actually used right now)
* Filtering the 10 million annoying astropy warnings


AnthonyHorton (approved):
Looks good. FWIW updates to astropy and/or numpy seem to have stopped the flood of warnings now but doesn't hurt to keep that filter in there.

PR #103, opened 2018-12-01
Small updates to dome

wtgee:
More dome changes are on the way but I accidentally was working in my `develop` branch and didn't want to disentangle so am just submitting here. :)

These are the updates that existed on the Mac from the latest run.


bazkiaei (approved):

conf_files/huntsman.yaml
```diff
@@ -30,14 +30,11 @@ directories:
     resources: POCS/resources/
     targets: /var/huntsman/huntsman-pocs/conf_files
     webcam: webcams
-cameras:
```
AnthonyHorton, line 33: Removing the camera section from the default config file seems to have broken the tests. Don't think it should, though.

AnthonyHorton (changes requested):
Tidying of musca.py is fine, but need to fix the tests.

conf_files/huntsman.yaml
wtgee: I'll get this fixed.  Issue here https://github.com/panoptes/POCS/issues/733

wtgee:
Closing in favor of #106

PR #111, opened 2019-10-18
Update camera properties and fix test suite

AnthonyHorton:
Update temperature control related camera property names in the Pyro camera classes to match those now used in POCS camera classes, and fix all the broken tests.

### DIFF
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -461,7 +461,7 @@ class CameraServer(object):
 
     @target_temperature.setter
     def target_temperature(self, target):
-        self._camera.target_temperature = set_point
+        self._camera.target_temperature = target
 
     @property
     def cooling_enabled(self):
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -43,48 +43,48 @@ class Camera(AbstractCamera):
         return self._serial_number
 
     @property
-    def ccd_temp(self):
+    def temperature(self):
         """
         Current temperature of the camera's image sensor.
         """
-        return self._proxy.ccd_temp * u.Celsius
+        return self._proxy.temperature * u.Celsius
 
     @property
-    def ccd_set_point(self):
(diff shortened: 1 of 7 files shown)


wtgee (approved):
:+1: Sorry I missed these yesterday. They weren't failing because the tests are just designed to skip them upon fail. Could probably clean that up in some fashion but more work than it is worth.

AnthonyHorton:
>  Sorry I missed these yesterday. They weren't failing because the tests are just designed to skip them upon fail. Could probably clean that up in some fashion but more work than it is worth.

https://github.com/panoptes/POCS/pull/909 enables those tests to be run, by using the SDK camera simulator instead of the DSLR one.

AnthonyHorton:
Tests are now only failing because it's trying to use the SDK camera simulators, which won't work until https://github.com/panoptes/POCS/pull/909 gets merged. I have the test suite running successfully locally.

huntsman/camera/pyro.py
```diff
@@ -450,30 +450,30 @@ def readout_time(self):
...
-        temperature = self._camera.ccd_set_point
+    def target_temperature(self):
+        temperature = self._camera.target_temperature
         return temperature.to(u.Celsius).value
 
-    @ccd_set_point.setter
-    def ccd_set_point(self, set_point):
-        self._camera.ccd_set_point = set_point
+    @target_temperature.setter
+    def target_temperature(self, set_point):
```
wtgee: Not a big deal, but for consistency.

```suggestion
    def target_temperature(self, target):
```

huntsman/camera/pyro.py
```diff
@@ -450,30 +450,30 @@ def readout_time(self):
...
+        self._camera.target_temperature = set_point
```
wtgee:
```suggestion
        self._camera.target_temperature = target
```

wtgee (approved):
LGTM.

Unfortunately some of this would also change again when/if docker becomes default.

PR #112, opened 2019-10-21
Don't perform a system exit if no local cameras are found.

wtgee:
Note: this will still show a `TERMINATING: No cameras available. Exiting.` warning. Should get rid of that too as it doesn't really exit.

Really with the cameras separated from the Observatory in POCS that error should no longer be existing so perhaps I will just make the fix there instead.

Update: https://github.com/panoptes/POCS/pull/913


AnthonyHorton (approved):
LGTM.

wtgee:
I might not merge this in favor of https://github.com/panoptes/POCS/pull/913.

wtgee:
This was fixed by the upstream PR mentioned above. Closing for now.

PR #116, opened 2019-10-23
Add Pyro Focuser class

### LINKED ISSUE #115: HuntsmanObservatory doesn't focus remote cameras
HuntsmanObservatory skips autofocus on all the distributed cameras because it thinks they don't have a focuser.

AnthonyHorton:
Adds a Pyro `Focuser` class to provide a functional `.focuser` attribute for the client side Pyro `Camera` class in order to enable direct access to remote focusers and fix #115.

### DIFF
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -221,11 +221,11 @@ class Camera(AbstractCamera):
                                                     base_name=base_name,
                                                     dark=bool(dark),
                                                     *args,
-                                                    **kwargs)
-        # Tag the file transfer on the end.
-        exposure_result = exposure_result.then(self._file_transfer, dir_name)
-        # Tag empty directory cleanup on the end & keep future result to check for completion
-        exposure_result = exposure_result.then(self._clean_directories)
+                                                    **kwargs) \
+            .then(self._file_transfer, dir_name) \
+            .then(self._clean_directories)
+
+        exposure_result
 
         # Start a thread that will set an event once exposure has completed
         exposure_thread = Timer(interval=seconds + self.readout_time,
@@ -437,18 +437,18 @@ class Camera(AbstractCamera):
         if future_result.wait(timeout):
             try:
                 result = future_result.value
-            except Exception as e:
-                self.logger.debug("Problem in wait: {}".format(e))
-                result = True
-                self.logger.debug("Setting result to True anyway")
-        else:
-            msg = "Timeout while waiting for {} on {}".format(name, self.name)
-            warn(msg)
-            self.logger.error(msg)
-            result = False
+            except Exception as err:
+                msg = f"Problem while waiting for {name} on {self.port}: {err}"
+                raise error.PanError(msg)
+            finally:
+                if event is not None:
+                    event.set()
 
-        if event is not None:
-            event.set()
+        else:
+            if event is not None:
+                event.set()
+            msg = "Timeout while waiting for {} on {}".format(name, self.port)
+            raise error.Timeout(msg)
 
         return result
 
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -175,7 +175,6 @@ class Camera(AbstractCamera):
                       filename=None,
                       dark=False,
                       blocking=False,
-                      timeout=None,
                       *args,
                       **kwargs):
         """
@@ -187,8 +186,6 @@ class Camera(AbstractCamera):
             dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
             blocking (bool, optional): If False (default) returns immediately after starting
                 the exposure, if True will block until it completes.
-            timeout (u.second, optional): Length of time beyond the length the exposure to wait
-                for exposures to complete. If not given will wait indefinitely.
 
         Returns:
             threading.Event: Event that will be set when exposure is complete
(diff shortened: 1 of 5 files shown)


huntsman/camera/pyro.py
```diff
@@ -92,13 +93,13 @@ def connect(self):
         """
         (re)connect to the distributed camera.
         """
-        self.logger.debug('Connecting to {} on {}'.format(self.name, self._uri))
+        self.logger.debug('Connecting to {} at {}'.format(self.port, self._uri))
```
wtgee, line 123: You're setting the `self.port` to the uri on Line 32 so here wouldn't these both be the same thing? Seems like you want to keep it at `name`.

https://github.com/AstroHuntsman/huntsman-pocs/blob/669f8831b4eba76d49e90b55e6fe3b121c93751c/huntsman/camera/pyro.py#L32

huntsman/camera/pyro.py
```diff
@@ -118,12 +119,18 @@ def connect(self):
...
         self._connected = True
         self._serial_number = uid
+        self.name = self._proxy.name
         self.model = self._proxy.model
         self._file_extension = self._proxy.file_extension
         self._readout_time = self._proxy.readout_time
         self._filter_type = self._proxy.filter_type
         self.logger.debug("{} connected".format(self))
 
+        if self._proxy._has_focuser:
```
wtgee: Should this be the private variable or public property? I'm confusing myself to how this is set up.
```suggestion
        if self._proxy.has_focuser:
```

huntsman/focuser/pyro.py
```diff
@@ -0,0 +1,82 @@
...
+class Focuser(AbstractFocuser):
+    """ Class representing the client side interface to the Focuser of a distributed camera. """
+    def __init__(self,
+                 name='Pyro Focuser',
+                 model='pyro',
+                 camera=None):
+
+        if not isinstance(camera, Camera):
+            msg = f"camera must be instance of huntsman.camera.pyro.Camera, got {type(camera)}."
+            raise ValueError
```
wtgee:
```suggestion
            raise ValueError(msg)
```

huntsman/focuser/pyro.py
```diff
@@ -0,0 +1,82 @@
...
+##################################################################################################
+
+    def connect(self):
+        # Pyro4 proxy to remote huntsman.camera.pyro.CameraServer instance.
+        self._proxy = self.camera._proxy
+        self.name = self._proxy.focuser_name
+        self.model = self._proxy.focuser_model
+        self.port = self.camera.port
+        self._serial_number = self._proxy.focuser_uid
+        self.logger.debug(f"{self} connected.")
```
wtgee, line 145: Does `self` come through as this Pyro object? Do we need a `__str__` method or is it fine to just see what ever is default?

wtgee (approved):
LGTM although I'll admit I'm a little confused as to how the camera and focus interact with each other via the proxy. Just need to spend more time looking at the whole I think.

AnthonyHorton:
@wtgee It is a bit convoluted. Basically I'm trying to produce the exact same interface for a distributed `Camera` _and_ its `.focuser` attribute as a local camera has, but because of the way Pyro4 works I'm doing it using a single proxy connection from the client side Pyro `Camera` class to the remote `CameraServer`. Consequently the client side Pyro `Focuser` class needs to use the proxy connection from the client side Pyro `Camera`, and the `CameraServer` needs a whole bunch more properties and methods that expose the interface of the `.focuser` of the remote `Camera`.

huntsman/focuser/pyro.py
AnthonyHorton: `self` here is the client side Pyro `Focuser` class, which inherits from `AbstractFocuser`, which has a `__str__` method which does something sensible.

Those inherited `__str__` methods are why I was fiddling around with the `name` and `port` attributes of the Pyro `Camera` class. I decided I'd like the string representation to be closer to what it would be for an equivalent local camera.

huntsman/camera/pyro.py
AnthonyHorton: Public property. Of a private variable. `self._proxy` is the Pyro4 connection to the remote `CameraServer`. `CameraServer.has_focuser` is a public property because it's supposed to be accessible remotely.

AnthonyHorton:
@wtgee That escalated quickly...

I have tried to update and complete the client side Pyro `Camera` interface so that it is (almost) identical to the local camera interface in the latest version of POCS, and also added a client side `Focuser` interface so that Pyro `Camera`'s `.focuser` attribute can be used to directly control the remote focuser (almost) exactly as if it were a local focuser. I have also updated the camera tests to (almost) match those currently in POCS.

wtgee (approved):
Looks good. But ugh. Seems like a lot of work-around code and that there should be some easier solution, but I'm not sure what that would be.

AnthonyHorton:
@wtgee Yeah, it's not pretty. So much boilerplate code, which Pyro is supposed to minimise.

There a few reasons for the the clunkiness. One is that I'm trying to do this for a whole bunch of existing camera, focuser, etc. classes that I want to leave untouched, instead of writing new classes that are designed to be distributed. I'm also trying to reproduce the exact same interface as those existing classes on the client side (i.e. distributed objects are a drop in replacement for a local one). Pyro itself then compounds these limitations because it is designed to expose the attributes of a remote object, but it doesn't appear to have any mechanism to directly expose the attributes of the attributes of a remote object, and so on.

These factors are what pushed me to make the camera server a wrapper around a camera object instead of being a camera object itself, and to add client side wrapper classes to reproduce the original interfaces. These wrappers on both server and client side then need access methods for all of the properties and methods of the camera and focuser classes, hence all the boilerplate.

Given the freedom to made sweeping, breaking changes (and lots more time!) I could redesign things to work directly with Pyro proxies and do away with the wrapper classes entirely, eliminating almost all of the repetitive boilerplate code.

PR #118, opened 2019-10-25
Fix egain, bit_depth properties

AnthonyHorton:
Fixes retrieval of remote camera `egain` and `bit_depth` properties by correctly stripping the units before serialisation and then adding them back afterwards.


JAAlvarado-Montes (approved):
LGTM!

PR #121, opened 2019-11-07
reset watchdog thread added

fergusL:
quick change to hopefully keep the dome open

### DIFF
--- a/huntsman/dome/musca.py
+++ b/huntsman/dome/musca.py
@@ -136,18 +136,21 @@ class HuntsmanDome(AbstractSerialDome):
         """Periodically tell musca to reset watchdog timer
 
         """
-        # check to see if a dome closure has occured
-        while True:
+        # maximum number of loops before dome closure (correspond to 15hrs)
+        max_loopps = 186
+        for i in range(max_loops):
+            # check to see if a dome closure has occured
             if self._close_event().is_set():
                 self.logger.info('Keep dome open thread has detected a dome closure, ending thread.')
                 # if dome has closed, don't try to 'keep dome open'
-                return
+                break
             self.logger.info('Sending keep dome open command to musca.')
             self._write_musca(Protocol.KEEP_DOME_OPEN, 'Keeping dome open.')
             # wait slightly less than 5 minutes to make sure we reset the
             # dome closure timer before it initiates a dome closure
             self.logger.debug('Keep dome open thread sleeping for ~5 minutes.')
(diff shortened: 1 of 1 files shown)


huntsman/dome/musca.py
```diff
@@ -120,10 +123,31 @@ def open(self):
...
+            self.logger.info('Keep dome open thread has detected a dome closure, ending thread.')
+            # if dome has closed, don't try to 'keep dome open'
+            return
+        self.logger.info('Sending keep dome open command to musca.')
+        self._write_musca(Protocol.KEEP_DOME_OPEN, 'Keeping dome open.')
+        # wait slightly less than 5 minutes to make sure we reset the
+        # dome closure timer before it initiates a dome closure
+        self.logger.debug('Keep dome open thread sleeping for ~5 minutes.')
+        time.sleep(290)
+
```
AnthonyHorton: All of this needs to be in a `while True:` loop otherwise it will only send the keep dome open command once.

huntsman/dome/musca.py
```diff
@@ -120,10 +123,31 @@ def open(self):
...
 
+    def keep_dome_open(self):
+        """Periodically tell musca to reset watchdog timer
+
+        """
+        # check to see if a dome closure has occured
+        if self._close_event().is_set():
+            self.logger.info('Keep dome open thread has detected a dome closure, ending thread.')
+            # if dome has closed, don't try to 'keep dome open'
+            return
```
AnthonyHorton: Once all of this is an `while True:` block probably best to replace this `return` with ` break`.

AnthonyHorton (approved):
Spotted one bug, but approving on the assumption that you fix it and acknowledgement that we need to test with the real hardware asap.

huntsman/dome/musca.py
```diff
@@ -120,10 +123,35 @@ def open(self):
...
+
+        """
+        # maximum number of loops before dome closure (correspond to 15hrs)
+        max_loopps = 186
+        for i in range(max_loops):
+            # check to see if a dome closure has occured
+            if self._close_event().is_set():
+                self.logger.info('Keep dome open thread has detected a dome closure, ending thread.')
+                # if dome has closed, don't try to 'keep dome open'
+                break
```
AnthonyHorton, line 146: Now that the enclosing loop is `for` then this probably wants to go back to being a `return`... Otherwise you'll get that warning about maximum loops every time you close the dome normally.

PR #123, opened 2019-11-20
Huntsman sshfs

danjampro:
Modified take_exposure in pyro.camera and pyro.camera_server to use sshfs

### DIFF
--- a/conf_files/huntsman.yaml
+++ b/conf_files/huntsman.yaml
@@ -30,6 +30,7 @@ directories:
     resources: POCS/resources/
     targets: /var/huntsman/huntsman-pocs/conf_files
     webcam: webcams
+    sshfs_mount: /home/huntsman/Huntsmans-Pro
 cameras:
     auto_detect: False
     devices:
--- a/conf_files/huntsman.yaml
+++ b/conf_files/huntsman.yaml
@@ -30,7 +30,6 @@ directories:
     resources: POCS/resources/
     targets: /var/huntsman/huntsman-pocs/conf_files
     webcam: webcams
-    sshfs_mount: /home/huntsman/Huntsmans-Pro
 cameras:
     auto_detect: False
     devices:
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -590,7 +590,7 @@ class CameraServer(object):
                                    blocking=True,
                                    *args,
                                    **kwargs)
-        return None
+        return filename
 
     def autofocus(self, *args, **kwargs):
         if not self.has_focuser:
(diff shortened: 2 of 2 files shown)


conf_files/huntsman.yaml
```diff
@@ -30,6 +30,7 @@ directories:
     resources: POCS/resources/
     targets: /var/huntsman/huntsman-pocs/conf_files
     webcam: webcams
+    sshfs_mount: /home/huntsman/Huntsmans-Pro
```
wtgee: I'm not actually sure if this is true or not.

```suggestion
    sshfs_mount: /home/huntsman/Huntsman-Pro
```

huntsman/camera/pyro.py
```diff
@@ -575,20 +570,27 @@ def get_uid(self):
...
-        filename = os.path.join(os.path.abspath(self.config['directories']['images']),
-                                './',
-                                base_name)
+    def take_exposure(self, seconds, filename, dark, *args, **kwargs):
+        
+        #Make sure the filename is accessible on the SSHFS
+        basename = os.path.normpath(self.config['directories']['base'])
+        filename = os.path.normpath(filename)
+        if not filename.startswith(basename):
+            raise ValueError("Filename not accesible on SSHFS.")
```
wtgee: This is not really testing if it is accessible or not. Either the test or the message should change. `if os.path.exists(filename)`?

huntsman/camera/pyro.py
```diff
@@ -575,20 +570,27 @@ def get_uid(self):
...
         # Start the exposure and wait for it complete
         self._camera.take_exposure(seconds=seconds,
                                    filename=filename,
                                    dark=dark,
                                    blocking=True,
                                    *args,
                                    **kwargs)
-        # Return the user@host:/path for created file to enable it to be moved over the network.
-        return "{}@{}:{}".format(self.user, self.host, filename)
+        return None
```
wtgee: Should still return just the plain filename. 

```suggestion
        return filename
```

conf_files/huntsman.yaml
wtgee: Why not just use `images`?

wtgee (commented):
@danjampro I made some comments and then realized it seems like we could just use the `directories.images` value directly and not require most of these code changes. Is there any reason to not just set the config accordingly?

conf_files/huntsman.yaml
danjampro: Nice suggestion. I was worried that changing 'images' would impact other parts of the code - if it is only used here then that would be a good way to go.

huntsman/camera/pyro.py
danjampro: wouldn't os.path.exists(filename) always return False anyway because the file hasn't been written yet? Perhaps something more like if os.path.isdir(os.path.dirname(filename))... ?

conf_files/huntsman.yaml
danjampro: I used "Huntsmans-Pro" for the mount point  because apparently that is its name (according to the google doc with the IP addresses)

wtgee (approved):
LGTM!

PR #125, opened 2019-11-27
huntsman/camera/pyro.py

danjampro:


huntsman/camera/pyro.py
```diff
@@ -570,11 +573,17 @@ def get_uid(self):
...
         return self._camera.uid
 
-    def take_exposure(self, seconds, base_name, dark, *args, **kwargs):
+    def take_exposure(self, seconds, base_name, dark, dir_name=None, *args,
+                      **kwargs):
         
         #Specify the full filename
-        filename = os.path.join(os.path.abspath(
+        #This uses the camera server's "images" directory 
+        if dir_name is None:
```
wtgee, line 581: I'm tempted to say that we should check if `base_name` is a relative or absolute path and only prepend the `images` directory if a relative directory is passed. That gets tricky though. Still, I think it would need to be well documented (and is a change in behaviour?) that the images will always be relative to the `images` directory.

wtgee (approved):
:+1:

huntsman/camera/pyro.py
danjampro: I don't think it is a change in behaviour because the previous version contained "filename = os.path.join(os.path.abspath(self.config['directories']['images']),'./',base_name)" in CameraServer.take_exposure. This defaults to that behaviour unless dir_name is not None.

PR #126, opened 2019-11-30
Adding tags to states and fixing IERS

### LINKED ISSUE #124: Not transitioning from ready state to calibrating (?)
When starting up huntsman before twilight, it seems we never make the transition from ready into calibrating (to take flats). Instead POCS ends up waiting until the sun is below -18 degree and then tries to take flats and concludes it is too dark.
This is probably due to a change to the `Observatory.is_dark()` method from POCS, where it now has a kwarg for the horizon level. By default this is `observe` which corresponds to -18 degree sun altitude. The two other options are `flat` and `focus`. This `is_dark()` method gets called by the `POCS.is_safe()` method so there should just be a `is_safe()` or `is_dark()` call somewhere that needs to have `horizon='flat'` passed to it. Then POCS should transition from the ready state to the calibrating state as desired.

wtgee:
* Adding state tags for safety and twilight.
  * Closes #124 after panoptes/POCS#930
* Use the development version of astroplan and the latest version of astropy so that we have respect for the IERS urls.

### DIFF
--- a/resources/state_table/huntsman.yaml
+++ b/resources/state_table/huntsman.yaml
@@ -3,19 +3,26 @@ name: huntsman
 initial: sleeping
 location: states
 states:
-    - parked
-    - sleeping
-    - housekeeping
-    - ready
-    - calibrating
-    - scheduling
-    - preparing
-    - slewing
-    - focusing
-    - observing
-    - analyzing
-    - dithering
-    - parking
+    parked:
+        tags: always_safe
+    sleeping:
+        tags: always_safe
+    housekeeping:
+        tags: always_safe
+    ready:
+        tags: always_safe
+    calibrating:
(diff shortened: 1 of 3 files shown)


resources/state_table/huntsman.yaml
```diff
@@ -3,19 +3,26 @@ name: huntsman
...
+        tags: always_safe
+    ready:
+        tags: always_safe
+    calibrating:
+        tags: at_twilight
+    scheduling:
+        tags: always_safe
+    preparing:
+    slewing:
+    focusing:
```
AnthonyHorton: IIRC the plan was to enter `focusing` before astronomical twilight too, after `calibrating`.

AnthonyHorton (approved):
LGTM.

PR #127, opened 2019-12-02
Auto sshfs script

danjampro:
Added a script to mount the SSHFS directory and create necessary symlink. Would be good to get the remote directory & local mount point directly from the config file at some point.

### DIFF
--- a/scripts/auto_sshfs_mount.py
+++ b/scripts/auto_sshfs_mount.py
@@ -46,11 +46,14 @@ def AttemptUnmount(local):
 if __name__ == '__main__':
     
     #Specify the mount point as ~/Huntsmans-Pro (for huntsman user)
-    local = '/home/huntsman/Huntsmans-Pro'
+    home = os.path.expandvars('$HOME')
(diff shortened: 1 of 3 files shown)


scripts/auto_sshfs_mount.py
```diff
@@ -0,0 +1,84 @@
...
+#==============================================================================
+
+if __name__ == '__main__':
+    
+    #Specify the mount point as ~/Huntsmans-Pro (for huntsman user)
+    local = '/home/huntsman/Huntsmans-Pro'
+    
+    #Specify the remote directory
+    dirs = load_config()['directories']
+    remote = f"huntsman@192.168.80.101:{dirs['base']}"
```
fergusL: I think this doesn't matter because it isnt our external ip address(?)... but we have had an issue before with an ip going into the repos commit history/not being easily able to remove it.

fergusL (commented):
LGTM,

scripts/auto_sshfs_mount.py
danjampro: Good point. I've updated it so that the IP is loaded from the local config file.

PR #128, opened 2019-12-09
Fix exposure

danjampro:
Quick fix for the (currently failing) test_exposure:

- Overrides is_ready in pyro.camera

- Checks is_exposing before any thing else in is_ready. This allows is_ready to behave as expected when an exposure is already in progress by bypassing the rest of the check. 

- For some reason, the other checks seemed to cause the is_ready method to wait until the exposure was completed previously. I do not have a clear idea about why this is, but is_temperature_stable may be the culprit...

### DIFF
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -212,11 +212,14 @@ class Camera(AbstractCamera):
 
         # Start the exposure
         filename = os.path.abspath(filename) 
-        base_name = os.path.split(filename)[-1]
+        dir_name, base_name = os.path.split(filename)
+        
         self.logger.debug(f'Taking {seconds} second exposure on {self.name}: {base_name}')
+        
         # Remote method call to start the exposure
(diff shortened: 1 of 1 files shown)


fergusL (approved):
I'm not familiar with the camera control code but LGTM

wtgee (approved):
LGTM although I would say to add some logging. Here we would just find out that camera is not ready but wouldn't know why. Some `debug` statements before each possible `return` might be helpful.

danjampro:
> LGTM although I would say to add some logging. Here we would just find out that camera is not ready but wouldn't know why. Some `debug` statements before each possible `return` might be helpful.

Done.

AnthonyHorton:
Very late to the party here, but it's bugging me that we have to duplicate so much `is_ready` code here. I think this is mostly a consequence of the decision to make the `take_exposure` method of the `CameraServer` be blocking, otherwise you'd just pass the `is_ready` call through to the remote camera.

PR #130, opened 2019-12-12
Dark collection state

JAAlvarado-Montes:
Hi all,

This is the first PR for the dark-collection mode of Huntsman, saved as `states/huntsman/taking_darks.py`. I based this mode on the housekeeping.py mode of Huntsman, and the modified scripts are `scheduler/observation.py` and `huntsman/observatory.py`. For the first one a new class `DarkObservation` has been included, and for the second one a new method was created `take_dark_fields` and a new private method called `_create_dark_observation`. All the new definitions have been based on previously existent code within the same scripts. I know there are many things to improve, so this is the first PR to start pulling off this new mode of Huntsman.

Any comment or suggestion is welcome.

Cheers,
Jaime.


states/huntsman/taking_darks.py
```diff
@@ -0,0 +1,26 @@
+def on_enter(event_data):
+    """Pointing State
+    Take dark fields
+    """
+    pocs = event_data.model
+
+    pocs.next_state = 'parking'
```
lspitler: Should the mount be parked before darks are taken?

states/huntsman/taking_darks.py
```diff
@@ -0,0 +1,26 @@
+def on_enter(event_data):
+    """Pointing State
```
lspitler:
```suggestion
    """
```

states/huntsman/taking_darks.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if pocs.observatory.mount.set_target_coordinates(current_observation.field):
+
+            pocs.status()
+
+            # Wait until mount is parked
+            pocs.say("I'm going to take next dark")
+
+            pocs.next_state = 'parked'
+
```
lspitler, line 20: I think you need to call `take_dark_fields` somewhere here, right?

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
                 self.logger.debug('Waiting for dark-field image')
                 time.sleep(1)
 
+    def take_dark_fields(self,
+                         max_exptime=60.,
```
lspitler: Why do you have this parameter?

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+                         camera_list=None,
+                         max_num_exposures=10,
```
lspitler: this parameter isn't used:
```suggestion
                         
```

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+            max_num_exposures (int, optional): Maximum number of darks to take
+            *args (TYPE): Description
+            **kwargs (TYPE): Description
+
+        """
+
+        image_dir = self.config['directories']['images']
+
+        dark_obs = self._create_dark_observation()
+        exp_times = {cam_name: [1. * u.second] for cam_name in camera_list}
```
lspitler: Looks like you're hard-coding 1 second exposures here for some reason. Shouldn't this be given as a arg to `take_dark_fields`?

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+
+        # Loop until conditions are met for dark fields
+        while True:
```
lspitler: How do you get out of this loop? I don't see a `break`

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+            # self.logger.debug("Slewing to parked-mount coords: {}".format(dark_obs.field))
+            # self.mount.set_target_coordinates(dark_obs.field)
```
lspitler: remove these lines?

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+
+                filename = "{}/darks/{}/{}/{}.{}".format(
+                    image_dir,
+                    camera.uid,
+                    dark_obs.seq_time,
+                    'dark_{:02d}'.format(dark_obs.current_exp_num),
+                    camera.file_extension)
+
+                # Take picture and get event
+                if exp_times[cam_name][-1].value < max_exptime:
```
lspitler: Don't think this is needed. So `max_exptime` is also not needed?

huntsman/observatory.py
```diff
@@ -386,6 +386,74 @@ def take_evening_flats(self,
...
+
+                    camera_events[cam_name] = {
+                        'event': camera_event,
+                        'filename': filename,
+                    }
+
+            # Block until done exposing on all cameras
+            while not all([info['event'].is_set() for info in camera_events.values()]):
+                self.logger.debug('Waiting for dark-field image')
+                time.sleep(1)
```
lspitler:
```suggestion
                time.sleep(wait_interval)
```

As in `pointing.py`

huntsman/observatory.py
```diff
@@ -469,3 +537,28 @@ def _create_flat_field_observation(self, alt=None, az=None,
         self.logger.debug("Flat-field observation: {}".format(flat_obs))
 
         return flat_obs
+
+    def _create_dark_observation(self, n_darks=9):
+        dark_coords = self.mount.get_current_coordinates().to_string('hmsdms')
+
+        self.logger.debug("Creating darks observation")
+        dark_field = Field('Darks', dark_coords)
+        dark_obs = DarkObservation(dark_field, exp_time=1. * u.second)
```
lspitler: Again, you're hardwiring in an exposure time.

AnthonyHorton:
@JAAlvarado-Montes When you get a chance please resolve the merge conflicts. The tests won't run until those are fixed.

lspitler (changes requested):
@JAAlvarado-Montes good first step. Some changes and questions for you here. Let me know when you've addressed them or you want to chat about them. Thanks

AnthonyHorton:
@JAAlvarado-Montes Have you tested this in simulation? Based on a quick look I don't think it would actually work. Your new state doesn't call the new HuntsmanObservatory method for taking darks at all, and if it did the HuntsmanObservatory would keep taking darks forever.

lspitler:
@JAAlvarado-Montes are you able to progress this so I can do another review before I leave on the 9th?

huntsman/observatory.py
wtgee: I also don't see a `break` :)

wtgee (changes requested):
Looks like this has mostly been hashed out and a copy of the diterhed except for the identified lack of `break` statement.

JAAlvarado-Montes:
Hi @lspitler, @wtgee, and @AnthonyHorton,

I have addressed all of the changes and suggestions you made for this PR. The three modified scripts are taking_darks.py (state), observatory.py containing the function `take_dark_fieds` and private method `_create_dark_observation`; and observation.py which has the subclass `DarkObservation` used by the aforementioned function. 

Something I'm not sure how to do (and that I haven't done yet) is how to decide when the taking_darks state should be triggered. I think the two important conditions to be met would be `if pocs.is_dark and not pocs.can_observe:...` but I'm not certain whether I'm missing something or if those two conditions are right. Also, where would these conditions go? I assume that they cannot go in the taking_darks state, so I think they should go to a different state (maybe observing.py?) that triggers the "dark" state, but I want to be sure before doing this step.

If you have some chance, please review this PR and let me know if you have any feedback. It will be really appreciated.

Cheers,
Jaime.

wtgee:
> Something I'm not sure how to do (and that I haven't done yet) is how to decide when the taking_darks state should be triggered. I think the two important conditions to be met would be `if pocs.is_dark and not pocs.can_observe:...` but I'm not certain whether I'm missing something or if those two conditions are right. Also, where would these conditions go? I assume that they cannot go in the taking_darks state, so I think they should go to a different state (maybe observing.py?) that triggers the "dark" state, but I want to be sure before doing this step.
> 
> If you have some chance, please review this PR and let me know if you have any feedback. It will be really appreciated.

I can explain how the state setup works and give a PR review later today.

JAAlvarado-Montes:
Great @wtgee, I'll be attentive to it. 

Cheers!

states/huntsman/taking_darks.py
```diff
@@ -35,7 +35,7 @@ def on_enter(event_data):
                 pocs.observatory.take_dark_fields(camera_list=cameras,
                                                   exptimes_list=exptimes)
 
-            pocs.next_state = 'parked'
-
     except Exception as e:
         pocs.logger.warning("Problem with preparing: {}".format(e))
+
+    pocs.next_state = 'parked'
```
wtgee: The mount has to be parked to do these darks, so the _previous_ state will be `parked`, but the `next_state` should probably be `housekeeping` (or `sleeping`).

wtgee: Following up on this, that also means that the logic to get **to** the `taking_darks` state should go in `parked`. 

It should be changed to something like:

```py
# states/huntsman/parked.py

# Psuedo-code below
if is_dark and weather_is_bad:
    pocs.next_state = 'taking_darks'
else:
    pocs.next_state = 'housekeeping'
```

It might be more complex than that, depending.

Alternatively, we could consider the act of taking darks to be part of `housekeeping` itself, but we probably best to keep it a little separate.

states/huntsman/parked.py
```diff
@@ -5,4 +5,8 @@ def on_enter(event_data):
     pocs.say("I'm parked now. Phew.")
 
     pocs.say("Cleaning up for the night!")
-    pocs.next_state = 'housekeeping'
+
+    if pocs.is_dark and pocs.weather_is_bad:
```
wtgee:
```suggestion
    if pocs.is_dark(horizon='flat') and pocs.is_weather_safe is False:
```

We don't have a `weather_is_bad`, I was just being lazy in my earlier comment (that's why it said it was psuedo-code).

JAAlvarado-Montes: OK, thanks for the input, I'll change this

JAAlvarado-Montes:
Hi @AnthonyHorton, @wtgee, and @danjampro,

I've modified the taking_darks.py script to have a `while` loop that meets the two necessary conditions for taking darks. However, I'd like you to give me some input on this part as I'm not sure whether the syntax I'm following is correct. I didn't put an explicit `break` statement because as far as I'm concerned that should be implicit(?). I mean, once those two conditions are not met anymore the `while` loop should be broken automatically and an `Exception` raised (which is what I did). After that `Exception` the `housekeeping` state is triggered.

Once I address this, I want to continue and do the required test for this state. Question: Should that test be placed in `huntsman/tests`? Or should it live in a different location? I ask since I don't see any other tests regarding huntsman states.

Thanks in advance for taking some time to review this, I appreciate it.

Cheers,
Jaime.

huntsman/tests/test_huntsman.py
```diff
@@ -180,6 +180,17 @@ def test_default_lookup_trigger(pocs):
     assert pocs._lookup_trigger() == 'parking'
 
 
+def test_darks_collection(pocs):
+    pocs.state = 'observing'
+
+    pocs.is_dark(horizon="flat") is True
```
danjampro:
```suggestion
    pocs.is_dark(horizon="flat") = True
```

huntsman/tests/test_huntsman.py
```diff
@@ -180,6 +180,17 @@ def test_default_lookup_trigger(pocs):
...
+    pocs.is_weather_safe is False
```
danjampro:
```suggestion
    pocs.is_weather_safe = False
```

huntsman/tests/test_huntsman.py
danjampro: Are you trying to setup the conditions for the dark state here? If so, I think = should be used.

states/huntsman/taking_darks.py
```diff
@@ -0,0 +1,42 @@
...
+                    if exp not in exptimes:
+                        exptimes.append(exp)
+
+                if len(cameras) > 0:
+                    pocs.say("Starting dark fields")
+                    pocs.observatory.take_dark_fields(camera_list=cameras,
+                                                      exptimes_list=exptimes)
+
+    except Exception as e:
+        pocs.logger.warning("Conditions for dark fields are not met: {}".format(e))
```
danjampro: I'm not sure the error message is appropriate. An exception won't be raised if the while condition is not met. Maybe:
```suggestion
        pocs.logger.warning("Problem encountered while taking darks: {}".format(e))
```

states/huntsman/taking_darks.py
```diff
@@ -0,0 +1,42 @@
...
+                for target in pocs.scheduler.fields_list:
+                    exp = target["exptime"].value
+                    if exp not in exptimes:
+                        exptimes.append(exp)
+
+                if len(cameras) > 0:
+                    pocs.say("Starting dark fields")
+                    pocs.observatory.take_dark_fields(camera_list=cameras,
+                                                      exptimes_list=exptimes)
+
```
danjampro, line 20: I think there should be a break statement in here, otherwise (e.g. on a rainy night) we would end up with 12 hours worth of darks. Either that, or you could change the `while` statement to an `if` statement e.g.:

```
if pocs.is_dark(horizon="flat") and pocs.is_weather_safe is False:
    <code here>
else:
    pocs.logger.info("Conditions for dark fields are not met")
```

huntsman/observatory.py
```diff
@@ -386,6 +386,78 @@ def take_evening_flats(self,
...
+                    dark_obs.seq_time,
+                    'dark_{:02d}'.format(dark_obs.current_exp_num),
+                    camera.file_extension)
+
+                # Take picture and get event
+                camera_event = camera.take_observation(
+                    dark_obs,
+                    fits_headers,
+                    filename=filename,
+                    exptime=exptimes[cam_name][-1]
```
danjampro: Should this not be a loop over exposure times? Why is `exptimes[cam_name][-1]` used?

huntsman/observatory.py
```diff
@@ -386,6 +386,78 @@ def take_evening_flats(self,
                 self.logger.debug('Waiting for dark-field image')
                 time.sleep(1)
 
+    def take_dark_fields(self,
+                         wait_interval=5,
+                         camera_list=None,
+                         exptimes_list=[],
```
danjampro: What happens if `exptimes_list` is empty? Do / should we have a predefined sequence of dark exposure times?

danjampro (changes requested):
Hi @JAAlvarado-Montes, I've added a few questions, suggestions and discussion points to the PR. I have two main concerns at the moment:

1) It doesn't seem that you are looping over the exposure times in `exptimes_list`, but rather doing one exposure using the last entry.

2) There is no way the dark collection state will stop if it is dark and there is bad weather. I think this will lead to far too many darks being taken on some nights (and will probably cause your test to hang indefinitely). I'm not sure the `try, while, except` structure you have used in `on_enter` will do what you intended it to.

huntsman/tests/test_huntsman.py
JAAlvarado-Montes: Yes, I am setting up the conditions for the dark state. I tried using `=` but got an error which says "can't assign to function call". However, I've changed the way to set up (simulate) those conditions. Please let me know what you think about that. Thanks!

huntsman/tests/test_huntsman.py
JAAlvarado-Montes: Same here (see the previous answer).

states/huntsman/taking_darks.py
JAAlvarado-Montes: Changed this to the suggested message. However, it's been moved to the `parked` state where the conditions for triggering `taking_darks` are.

states/huntsman/taking_darks.py
JAAlvarado-Montes: I noticed that I put the conditions to trigger the `taking_darks` state inside the `parked` state using an `if` statement as you suggested. For that reason, I have removed them from the `taking_darks` state since if this state is triggered it's because the conditions were already met in `parked`, so it's not necessary to repeat them inside `taking_darks`. `taking_darks` now has only a `try` and `except` statement.

huntsman/observatory.py
JAAlvarado-Montes: The [-1] was a typing error. Thanks for this!

huntsman/observatory.py
JAAlvarado-Montes: Well, I reckon there should always be some targets in the config file, and the exposure times are taken from there to fill the list that is passed into the function `take_dark_fields`. So the only way for that list to be empty is that the targets config file is totally empty. If that happens, then `exptimes_list` would be empty. But I'm not sure if that's feasible since we always start the night with some targets in that file (and hence their exposure times), don't we? 
If there's a possibility that such a file be empty for any reason, then as you pointed out we should have a predefined list of exposure times. It totally depends on the config file for the targets.

JAAlvarado-Montes:
Hello @AnthonyHorton, @danjampro, and @wtgee,

This PR is ready for a "final" review. On the one hand, I have written two tests to `test_huntsman.py`, namely  `test_entering_darks_state` and `test_darks_collection_simulator`. The former is to test the conditions for entering the `taking_darks` state, and the latter is for testing the method that takes darks. For testing I'm using the docker test image environment and both tests are passing without any conflict. `test_entering_darks_state` simply simulates the conditions needed for taking darks, which are when the sky is dark and the weather is not safe. `test_darks_collection_simulator` does simply the same than the state would do when taking darks, but using simulated cameras. It searches into the `pocs.observatory.scheduler.fields_list` to collect all of the exposure times that have been used before when observing, and create a list of exposure times (without repeating) which will be parsed to the `take_dark_fields` method in `observatory.py`

On the other hand, the `take_dark_fields` method in `observatory.py` is designed so that given a list of exposure times it will take a sequence of 10 darks per exposure time (i.e. by default, but can be modified) exposing all of the cameras at the same time and waiting for all of them to finish before taking the next exposure in the sequence. Once a sequence is done, it continues with the next exposure time on the list and so on.

I have tested `test_darks_collection_simulator` for different numbers of exposure times in `pocs.observatory.scheduler.fields_list` and it is working perfectly.  However, your recommendations are very welcome and I would be glad to add any change you consider important if you spot some things I have not, so I appreciate your input.

Cheers,
Jaime.

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+
+    def __init__(self, position, *args, **kwargs):
+        """
+        Args:
+            position (str): Center of field, can be anything accepted by
+                `~astropy.coordinates.SkyCoord`.
+        """
+        # Create the observation
+        dark_coords = position.to_string('hmsdms')
+        dark_field = Field('Dark Field', dark_coords)
```
wtgee: You should be able to just pass the `position` directly

```suggestion
        dark_field = Field('Dark Field', position)
```

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+
+        # Set initial list to original values
+        self._exptime = listify(self.exptime)
+        self._field = listify(self.field)
+
+        self.extra_config = kwargs
+
+    @property
+    def exptime(self):
+        exptime = self._exptime[self.exposure_index]
```
wtgee: This could use a docstring with some explanation about how this `exptime` is different.

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
 
         # Specify directory root for file storage
         self._directory = os.path.join(self._image_dir, 'flats')
+
+
+class DarkObservation(Observation):
```
wtgee: This whole class could probably go in its own file.

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+
+    @exptime.setter
+    def exptime(self, values):
+        assert all(t > 0.0 for t in listify(values)), \
+            self.logger.error("Exposure times (exptime) must be greater than 0")
+
+        self._exptime = listify(values)
+
+    @property
+    def field(self):
```
wtgee: Same about the docstring.

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+
+        self._field = listify(values)
+
+    @property
+    def exposure_index(self):
+        _exp_index = 0
+        with suppress(AttributeError):
+            _exp_index = self.current_exp_num % len(self._exptime)
+
+        return _exp_index
```
wtgee: We might want to build in an actual exposure index that increments when taking pictures but I guess I don't see any need.

But `_exp_index` doesn't need the `_` ;)

huntsman/scheduler/observation.py
```diff
@@ -113,3 +113,73 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+
+    def __str__(self):
+        return "DarkObservation: {}: {}".format(self._field, self._exptime)
```
wtgee: Just a reminder you can get in the habit of:

```suggestion
        return f"DarkObservation: {self._field}: {self._exptime}"
```

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+        """Take dark fields for each camera with all the exposure times
+           in the targets list.
+
+        Args:
+            exptimes_list (list): List of exposure times for darks
+            wait_interval (float, optional): Time in between dark exposures
+            camera_list (list, optional): List of cameras to use for darks
+            n_darks (int, optional): Number of darks to be taken per exptime
+            *args (TYPE): Description
+            **kwargs (TYPE): Description
```
wtgee: This needs to be cleaned up to match.

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+        """
+
+        if camera_names is None:
+            cameras_all = self.cameras
```
wtgee: `cameras_all` is a bad name because it is not necessarily always all cameras (such as in this line here).

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+        else:
+            cameras_all = {c: self.cameras[c] for c in camera_names}
+
+        self.logger.debug('Using cameras {}'.format(cameras_all))
+
+        image_dir = self.config['directories']['images']
+
+        self.logger.debug("Going to take {} dark fields for each of these exposure times {}".
+                          format(n_darks, exptimes))
```
wtgee: Good examples of the f-strings being better.
```suggestion
        self.logger.debug(f"Going to take {n_darks} dark fields for each of these exposure times {exptimes}")
```

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+        # List to check that the final number of darks is equal to the number
+        # of cameras times the number of exptimes times n_darks.
+        total_exposures = []
+
+        # Loop over cameras.
+        for exptime in exptimes:
+
+            self.status()  # Seems to help with reading coords
```
wtgee: I want to say let's remove this and try to remember in case we see weird things. But ideally we don't have a line we don't understand.

```suggestion
```

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+            start_time = utils.current_time()
+
+            if isinstance(exptime, u.Quantity):
+                exptime = exptime.to_value(u.second)
```
wtgee: Slightly more pythonic to just assume and ignore exception.
```suggestion
            with suppress(AttributeError):
                exptime = exptime.to_value(u.second)
```

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+            # Loop over exposure times for each camera.
+            for num in range(n_darks):
+
+                self.logger.debug('Darks sequence #{} of exposure time {}s'.
+                                  format(num, exptime))
```
```suggestion
                self.logger.debug(f'Dark sequence #{num} of exposure time {exptime}s')
```

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+                camera_events = dict()
+
+                # Take a given number of exposures for each exposure time.
+                for camera in cameras_all.values():
+
+                    self.logger.debug('Camera {} is exposing for {}s'.
+                                      format(camera.uid, exptime))
```
wtgee: This statement isn't quite true yet as we haven't yet started exposing.  Could also change to f-string.

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                    fits_headers = self.get_standard_headers(observation=dark_obs)
+                    # Common start time for cameras
+                    fits_headers['start_time'] = utils.flatten_time(start_time)
+
+                    filename = '{}/darks/{}/{}/{}.{}'.format(
+                        image_dir,
+                        camera.uid,
+                        dark_obs.seq_time,
+                        'dark_{:02d}_{}s'.format(num, int(exptime)),
+                        camera.file_extension)
```
wtgee: Better to use `os.path.join` here:

```suggestion
                    filename = os.path.join(
                        image_dir,
                        'darks',
                        camera.uid,
                        dark_obs.seq_time,
                        f'dark_{num:02d}_{exptime:.02f}s.{camera.file_extension}')
```

Although I'm not sure we want to continue to hard-code the `darks` directory.  Also the filename is a little weird with the exptime so you might want to alter to waht you want.

huntsman/observatory.py
```diff
@@ -316,6 +314,105 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+                    # Take picture and get event
+                    camera_event = camera.take_observation(
+                        dark_obs,
+                        fits_headers,
+                        filename=filename,
+                        exptime=exptime,
+                        dark=True,
+                        blocking=True
```
wtgee: I don't think you want to block here as it waits below.

```suggestion
                        blocking=False
```

wtgee (commented):
Looking good. I did a first quick pass because there are some changes in it and tests aren't passing right now. 

So sorry, probably not the "final" version quite yet. :wink:

huntsman/observatory.py
```diff
@@ -325,26 +326,22 @@ def take_dark_fields(self,
            in the targets list.
 
         Args:
-            exptimes_list (list): List of exposure times for darks
+            exptimes (list): List of exposure times for darks
             wait_interval (float, optional): Time in between dark exposures
```
danjampro: What are the units of `wait_interval`? Probably seconds - but good to be specific.

states/huntsman/parked.py
```diff
@@ -6,7 +6,7 @@ def on_enter(event_data):
 
     pocs.say("Cleaning up for the night!")
 
-    if pocs.is_dark(horizon="flat") and not pocs.is_weather_safe():
+    if pocs.is_dark() and not pocs.is_weather_safe():
         pocs.next_state = 'taking_darks'
```
danjampro, line 10: O.K. so we are using the state `taking_darks` to actually take the darks. The problem is I can't see this state listed in the state table (`resources/state_table/huntsman.yaml`), which needs to declare the specify the allowable transitions in/out of `taking_darks`.

huntsman/observatory.py
```diff
@@ -325,26 +326,22 @@ def take_dark_fields(self,
            in the targets list.
```
danjampro: This description is not quite right here because this function does not look up the exposure times in the target list. Instead, this function expects an `exptimes` argument. Maybe a better description would be something like:

"Take n_darks dark frames for each exposure time specified, for each camera."

huntsman/tests/test_huntsman.py
```diff
@@ -252,8 +252,6 @@ def test_entering_darks_state(pocs, db):
 
     pocs.state = 'taking_darks'
```
danjampro: This is not testing whether it can enter the `taking_darks` state (I don't actually think it can at the moment because there is no mention of `taking_darks` in the state table, see comment below).

So first you assert that the conditions are right to enter the dark state. So you need it to be a) dark, b) weather not safe and c) in the correct state to transition into `taking_darks`. It looks like this should be `parked` from the rest of your code, not `sleeping`. You will then need something like:

```
        assert pocs.next_state == "taking_darks"
        pocs.goto_next_state()
        assert pocs.state == "taking_darks"
``` 

for this to actually test whether it successfully enters the state. Finally, I think this test would be better suited to being in `tests/test_states.py` rather than `tests/test_huntsman.py`. I would suggest taking a look at some of the other tests in `tests/test_states.py` if this is not clear.

huntsman/tests/test_huntsman.py
```diff
@@ -252,8 +252,6 @@ def test_entering_darks_state(pocs, db):
...
 
-    # assert(pocs.state == 'parked')
-
 
 def test_darks_collection_simulator(pocs, tmpdir):
```
danjampro, line 236: - It would be nice to include a short description of what this test is trying to achieve.
- What is the purpose of `return False` at the end of this function?

huntsman/tests/test_huntsman.py
```diff
@@ -276,8 +274,6 @@ def test_darks_collection_simulator(pocs, tmpdir):
     exptimes_list = list()
     for target in pocs.observatory.scheduler.fields_list:
         exptime = target['exptime']
-        if not isinstance(exptime, u.Quantity):
-            exptime *= u.second
         if exptime not in exptimes_list:
             exptimes_list.append(exptime)
```
danjampro, line 260: I think you need to actually check whether the FITS files are actually created at the end of this test. You can use something like 
`assert(os.path.isfile(filename))`, or you could count the number of FITS files in the relevant directories etc.

danjampro (changes requested):
Hi @JAAlvarado-Montes, thanks for your work on this so far. I have two concerns that should be addressed before we merge this:

1) There is no way for the state machine to transition in/out of the `taking_darks` state because it is not in the state resource table (see comments below).

2) The test to check if the state machine successfully enters the `taking_darks` state does not work. If it did, it would pick up on the above. Simply changing the state like `pocs.next_state="taking_darks"` directly is not what we want to test, but rather we should check if the state machine naturally transitions to `taking_darks` given the right conditions. See below for some suggestions on this.

The other thing is that our tests now fail because of the Travis build (not related to your PR). @wtgee , thoughts on this?

huntsman/tests/test_huntsman.py
JAAlvarado-Montes: I'm mainly asserting two things now: 1) To check whether all of the created files are fits, and 2) verify that the number of created fits files equals the theoretical (expected) number of darks.

JAAlvarado-Montes:
Hello @danjampro, @wtgee, and @AnthonyHorton  

I have addressed all of @wtgee's and @danjampro's comments. As it is now, tests are passing and **we are indeed testing** that given the required conditions, namely bad weather and dark sky, the state machine goes from `parked` to `taking_darks`. I have added the `taking_darks` state to the states table along with its allowable transitions and migrated `test_entering_darks_state()` from `tests/test_huntsman.py` to `tests/test_states.py`

Please let me know what you think of this PR's current state, and as always, any thought or recommendation for moving it forward is very welcome.

Cheers,
Jaime A.

JAAlvarado-Montes:
Hello @danjampro, @wtgee, and @AnthonyHorton

I have addressed all of @wtgee's and @danjampro's comments. As it is now, tests are passing and we are indeed testing that given the required conditions, namely bad weather and dark sky, the state machine goes from parked to taking_darks. I have added the taking_darks state to the states table along with its allowable transitions and migrated test_entering_darks_state() from tests/test_huntsman.py to tests/test_states.py

Please let me know what you think of this PR's current state, and as always, any thought or recommendation for moving it forward is very welcome.

Cheers,
Jaime A.

resources/state_table/huntsman.yaml
```diff
@@ -49,6 +51,14 @@ transitions:
         source: parked
         dest: housekeeping
         trigger: clean_up
+    -
+        source: parked
+        dest: taking_darks
+        trigger: set_park
```
danjampro: I think the trigger to enter the `taking_darks` state should have its own trigger. Something like `trigger: take_darks`.

danjampro (commented):
@JAAlvarado-Montes  Looks good to me. I left one comment about changing the `trigger` for entering the new state. Once that's  implemented I will approve!

JAAlvarado-Montes:
Hi @danjampro,

I have addressed your comment as suggested. Please let me know whether you find it's ok, so I can close and merge this PR 👍 

Cheers,
Jaime.

danjampro (approved):
LGTM. Let's wait for either @AnthonyHorton of @wtgee to approve before the merge.

PR #131, opened 2019-12-15
NGAS push

danjampro:
Modified Camera in pyro.py to perform NGAS push following a successful call to take_exposure.

- By default, take_exposure does not attempt an NGAS push.

- By default, take_observation does attempt the NGAS push and specifies the NGAS filename as the collapsed filename defined in _setup_observation.

- Similar functionality also needs to be added for calibration exposures, but not autofocus ones.

Let me know what you think!

### DIFF
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -404,7 +404,7 @@ class Camera(AbstractCamera):
         # Start a thread that will set an event once autofocus has completed
         autofocus_event = Event()
         autofocus_thread = Thread(target=self._async_wait,
-                                  args=(autofocus_result, 'autofocus', autofocus_event, timeout))
+                args=(autofocus_result, 'autofocus', autofocus_event, timeout))
         autofocus_thread.start()
 
         if blocking:
@@ -508,7 +508,7 @@ class Camera(AbstractCamera):
             The local filename, the basename of which is the default NGAS
             filename.
         filename_ngas (str, optional):
-            The NGAS filename. If None, filename_ngas=basename(filename)
+            The NGAS filename. If None, auto-assign based on filename.
         ngas_push (bool, optional):
             Will attempt the push only if True. Else, does nothing!
         port (int, optional):
@@ -550,10 +550,8 @@ class Camera(AbstractCamera):
     
     #Might be a good idea to override _setup_observation to output filename
     #suitible for NGAS...
-    
-    """
     def take_observation(self, observation, headers=None, filename=None,
-                         **kwargs):
+                         filename_ngas=None, **kwargs):
         '''
         Take an observation.
         Gathers various header information, sets the file path, and calls
@@ -582,8 +580,23 @@ class Camera(AbstractCamera):
         exptime, file_path, image_id, metadata = self._setup_observation(
                                     observation, headers, filename, **kwargs)
         
+        #Define the NGAS filename as the collapsed filename?
+        if filename_ngas is None:
+            
+            #Remove the directory prefix specified in _setup_observation
+            prefix = os.path.normpath(observation.directory)
+            filename_ngas = os.path.normpath(file_path).split(prefix)[1]
+            
+            #Replace slashes by underscores
+            filename_ngas = filename_ngas[1:].replace(os.sep, '_')
+            
+            self.logger.debug(file_path)
+            self.logger.debug(filename_ngas)
+                    
+        #Take the exposure, passing the NGAS filename
         exposure_event = self.take_exposure(seconds=exptime,filename=file_path,
-                                            **kwargs)
+                                            filename_ngas=filename_ngas,
+                                            ngas_push=True, **kwargs)
 
         # Add most recent exposure to list
         if self.is_primary:
@@ -601,7 +614,7 @@ class Camera(AbstractCamera):
         t.start()
 
         return observation_event
-    """
+
 #==============================================================================
         
 @Pyro4.expose
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -216,7 +216,7 @@ class Camera(AbstractCamera):
             dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
             blocking (bool, optional): If False (default) returns immediately after starting
                 the exposure, if True will block until it completes.
-            ngas_push (bool, optional): Push file to NGAS database?
+            ngas_push (bool, optional): Push file to NGAS database? Default is False.
 
         Returns:
             threading.Event: Event that will be set when exposure is complete
@@ -499,8 +499,7 @@ class Camera(AbstractCamera):
         return result
     
     
-    def _NGASpush(self, filename, filename_ngas=None, ngas_push=True,
-                  port=7778):
+    def _NGASpush(self, filename, filename_ngas, ngas_push=True, port=7778):
         '''
         Parameters
         ----------
@@ -521,10 +520,14 @@ class Camera(AbstractCamera):
         '''
         if not ngas_push: #No need to do anything
             return True
-                
-        if filename_ngas is None:
-            filename_ngas = os.path.basename(filename)
         
+        #Make sure a valid filename is given
+        try:
+            assert(type(filename_ngas)==str)
+        except AssertionError as e:
+            self.logger.error('NGAS filename needs to be specified for NGAS push.')
+            raise e
+                                    
         #Get the IP address of the NGAS server
         ngas_ip = self.config['messaging']['ngas_server_ip']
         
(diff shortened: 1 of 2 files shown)


huntsman/camera/pyro.py
```diff
@@ -1,5 +1,7 @@
 import sys
 import os
+import requests
+import threading
```
AnthonyHorton: Already importing `Event` and `Timer` from `threading` in lines 6 & 7. Looks like you need to `from threading import Thread`.

huntsman/camera/pyro.py
```diff
@@ -211,6 +216,7 @@ def take_exposure(self,
             dark (bool, optional): Exposure is a dark frame (don't open shutter), default False
             blocking (bool, optional): If False (default) returns immediately after starting
                 the exposure, if True will block until it completes.
+            ngas_push (bool, optional): Push file to NGAS database? Default is False.
```
AnthonyHorton: Missing docstring for `filename_ngas` argument?

huntsman/camera/pyro.py
```diff
@@ -489,8 +497,126 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
+        -------
+        bool:
+            True if success, False if not.
+        '''
+        if not ngas_push: #No need to do anything
+            return True
+        
+        #Make sure a valid filename is given
+        try:
+            assert(type(filename_ngas)==str)
```
AnthonyHorton: `type(<object>) == <class/type>` is discouraged. It doesn't work for subclasses. Better to use `isinstance(<object>, <class/type>)` instead, but that can still end up inadvertently overly restrictive.

Best practice in Python is to use 'duck typing' (as in, "if it quacks like a..."), i.e. base checks on the behaviour of the object rather that explicit class/type checking. Here all you really require is that `filename_ngas` can be converted to a string in the f-string below, which will work for almost any Python object, and if it doesn't it will raise an exception there. I don't think you actually need this check at all.

huntsman/camera/pyro.py
```diff
@@ -489,8 +497,126 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
+    
+    
+    def _NGASpush(self, filename, filename_ngas, ngas_push=True, port=7778):
+        '''
+        Parameters
+        ----------
+        filename (str):
+            The local filename, the basename of which is the default NGAS
+            filename.
+        filename_ngas (str, optional):
```
AnthonyHorton, line 516: Not actually optional, either in the function definition or in the code below (assertion that it is a string).

huntsman/camera/pyro.py
```diff
@@ -489,8 +497,126 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
+            try: 
+                #Post the file
+                r = requests.post(url, files=files)
+                
+                #Confirm success
+                r.raise_for_status()
+                
+            except Exception as e: 
+                self.logger.debug('Error while performing NGAS push.')
+                self.logger.error(e)
```
AnthonyHorton: Purely a question of style, but I'd tend to combine these two as `self.logger.error(f"Error while performance NGAS push: {e}")`

huntsman/camera/pyro.py
```diff
@@ -489,8 +497,126 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
+                return False
 
+        return True
+    
+    
+    #Might be a good idea to override _setup_observation to output filename
+    #suitible for NGAS...
```
AnthonyHorton: Not sure I like this. Overriding this whole method from POCS just to change the formatting of the filename seems like a bad idea, maintenance wise. It means more code in huntsman-pocs that we need to manually keep in sync with upstream changes in POCS. We try to keep the amount of code in huntsman-pocs to the minimum, only including code that is too specific to Huntsman to include in POCS.

I think it would be preferable to just have the filename munging in the `_NGAS_push` method instead.

huntsman/camera/pyro.py
```diff
@@ -241,16 +247,18 @@ def take_exposure(self,
...
         exposure_result = self._proxy.take_exposure(seconds=seconds,
                                                     base_name=base_name,
                                                     dir_name=dir_name,
                                                     dark=bool(dark),
                                                     *args,
-                                                    **kwargs) 
-        exposure_result
+                                                    **kwargs
+                    #Once exposure is completed, push file to NGAS if necessary
+                    ).then(self._NGASpush, filename_ngas, ngas_push)
```
AnthonyHorton: This is too early to be pushing the file to NGAS. For an image that is taken via `take_observation()` the call to `take_exposure()` will be followed by a call to `process_exposure()`, which in turn calls `_process_fits()`, which adds the info about the `Observation` to the FITS headers with a call to `pocs.utils.fits.update_observation_headers` (at the moment that's all it does). If you push to NGAS here then the ingested files will be missing a lot of important metadata.

The best way to do this is probably to override `_process_fits`, calling the base class version then doing the NGAS push. That would limit NGAS pushes to only exposures started with `take_observation`, but I don't think that's a major problem.

huntsman/camera/pyro.py
```diff
@@ -600,7 +726,20 @@ def get_uid(self):
...
                       **kwargs):
-        
+        '''
+        Parameters
+        ----------
+        base_name (str):
+            The basename of the file.
+        dir_name (str, optional):
+            The directory in which to store the file. If None, will use
+            the images directory indicated by self.config.
```
AnthonyHorton: I like to point out that this `base_name`, `dir_name` split was only introduced to make the rsync file transfer work. Now that you've removed the need for that with the sshfs mount the directory name manipulation is just unnecessary faffing around. You should remove the code on the `Camera` side that splits up the `filename` and the code on the `CameraServer` side that replaces the `dir_name`, and have this `take_exposure` method just take a `filename` like the methods of the `Camera` classes do.

Deleting code is even better than writing code. We should do it every chance we get!

huntsman/camera/pyro.py
```diff
@@ -200,6 +203,8 @@ def take_exposure(self,
                       filename=None,
                       dark=False,
                       blocking=False,
+                      ngas_push=False,
+                      filename_ngas=None,
```
AnthonyHorton: If you're going to pass these options in to the class constructor shouldn't you be storing them as instance variables to use as defaults for the later function calls?

AnthonyHorton (commented):
Added a number of comments, most important of which is the need to move the pushing to later on (in `_process_fits`, not `take_exposure`).

huntsman/camera/pyro.py
```diff
@@ -489,8 +487,62 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
 
-
+    def _process_fits(self, file_path, info, *args, **kwargs):
+        '''
+        Override _process_fits, called by process_exposure in take_observation.
+        
+        The difference is that we do an NGAS push following the processing.
+        '''
+        #Call the super method
+        result = super()._process_fits(file_path, info, *args, **kwargs)
```
AnthonyHorton: `pocs.camera.camera.AbstractCamera._process_fits()` doesn't actually accept `*args` or `**kwargs` so if someone actually tried to pass in extra arguments here they'd get an error.

https://github.com/panoptes/POCS/blob/develop/pocs/camera/camera.py#L712

huntsman/camera/pyro.py
```diff
@@ -489,8 +487,62 @@ def _async_wait(self, future_result, name='?', event=None, timeout=None):
...
+        filename_ngas (str, optional):
+            The NGAS filename. If None, auto-assign based on filename.
+        port (int, optional):
+            The port of the NGAS server. Defaults to the TCP port.
+            
+        '''        
+        #Define the NGAS filename 
+        if filename_ngas is None:
+            extension = os.path.splitext(filename)[1] 
+            filename = f"{metadata['image_id']}{extension}"
```
AnthonyHorton, line 525: Huh. Yeah, that'll do if all we want is for the filename to be unique and somewhat identifiable. Might run into trouble down the track if we get high frame rate stuff going because of the low resolution of the timestamp in `image_id`, though. That will also be a problem for the on disc filenames too.

huntsman/camera/pyro.py
```diff
@@ -598,18 +650,18 @@ def get_uid(self):
...
+        Parameters
+        ----------
+        filename (str):
+            The filename of the exposure result.
+            
+        Returns
+        -------
+        str:
+            The full filename of the exposure output.
+        '''
```
AnthonyHorton, line 666: Returning the filename is another hangover from the rsync stuff. Doesn't do any harm, but equally we don't need it any more.

AnthonyHorton (approved):
Looks good. I suggested a few small changes, but then LGTM.

huntsman/camera/pyro.py
danjampro: Yes, I think that could be addressed by modifying the definition of `image_id` in the future.

danjampro:
Merging following approval.

Issue #132, opened 2019-12-16
add auto sshfs mount to camera server


lspitler:
Don't want to do manual step:
`python /huntsman-pocs/scripts/auto_sshfs_mount.py`

Instead should:
"The camera code doesn't know about the network mounts. It will just try to write to the Pi's filesystem wherever you tell it to. 
The mounting/demounting of the sshfs mount should really go into the camera server code.
Instead of being a separate manual step."

@AnthonyHorton @danjampro does this just go here: 
https://github.com/AstroHuntsman/huntsman-pocs/blob/develop/scripts/pyro_camera_server.py
?

danjampro:
Change implemented and tested

PR #133, opened 2019-12-16
sshfs mount in camera server

danjampro:
Camera server script now calls auto_sshfs_mount.MountSSHFS before starting the camera server.

From #132


lspitler (approved):
@danjampro LGTM. When this is approved, etc. Reset the Pis please.

AnthonyHorton:
All the tests that depend on the camera server generating an error in Travis CI (`TimeoutError: Timeout waiting for camera server to start`) so there's a bug here that needs addressing.

scripts/auto_sshfs_mount.py
```diff
@@ -12,7 +12,7 @@
 
 #==============================================================================
 
-def Mount(local, remote, ServerAliveInterval=20, ServerAliveCountMax=3):
+def _Mount(local, remote, ServerAliveInterval=20, ServerAliveCountMax=3):
```
AnthonyHorton: In Python convention CamelCase is for classes. Function and variable names should use lower case with underscores. Should be: `def _mount(local, remote, server_alive_interval=20, server_alive_count_max=3):`

AnthonyHorton: Is there a particular reason for making this function and attempt_unmount semi-private? That sort of thing makes more sense in a class than a script.

scripts/auto_sshfs_mount.py
```diff
@@ -29,7 +29,7 @@ def Mount(local, remote, ServerAliveInterval=20, ServerAliveCountMax=3):
         raise(e)
     
     
-def AttemptUnmount(local):
+def _AttemptUnmount(local):
```
AnthonyHorton: As with `_mount` this should be `_attempt_unmount`, and probably doesn't need to be private?

scripts/auto_sshfs_mount.py
```diff
@@ -41,10 +41,11 @@ def AttemptUnmount(local):
             pass
         
 #==============================================================================
-#==============================================================================
 
-if __name__ == '__main__':
+def MountSSHFS():
```
AnthonyHorton: `def mount_sshfs():`

scripts/auto_sshfs_mount.py
```diff
@@ -41,10 +41,11 @@ def AttemptUnmount(local):
...
+    '''
     
+    '''
     #Specify the mount point as ~/Huntsmans-Pro (for huntsman user)
     home = os.path.expandvars('$HOME')
     local = os.path.join(home, 'huntsman', 'Huntsmans-Pro')
```
AnthonyHorton: I'm still unclear why you mount as `~/Huntsman-Pro`, requiring a symlink, rather than just mounting directly at `$PANDIR/images`.

scripts/auto_sshfs_mount.py
```diff
@@ -57,11 +58,11 @@ def AttemptUnmount(local):
     
     #Check if the remote is already mounted, if so, unmount
     print(f'Attempting to unmount {local}...')
```
AnthonyHorton: print output has a tendency to get lost/redirected. Good to at least provide the option of writing to log file. For utility functions I've simply made `logger` an optional argument so that when called from POCS/huntsman-pocs classes you can just pass in `self.logger` and have the utility function write to the main log file.

danjampro:
@AnthonyHorton Yes, I think the cause is that the SSHFS mount fails because it is not set up to work with Travis. Since this is now part of the camera server script, that fails and causes functions requiring a camera to hang until timeout. I don't know enough about Travis to address this, yet, but I do not believe its an actual bug

scripts/auto_sshfs_mount.py
AnthonyHorton: The three functions (`mount_sshfs()`, `mount()` and `attempt_unmount`) should be moved to a file in the `huntsman/utils`.  Files in `scripts` should only contain the scripty bits, i.e. import statements, the `if __name__ == '__main__'`, perhaps some argument parsing, then a function call.

scripts/pyro_camera_server.py
```diff
@@ -5,11 +5,19 @@
...
+import os, sys
 import argparse
-
 from huntsman.utils.pyro import run_camera_server
 
 if __name__ == "__main__":
+    
+    #Mount the SSHFS 
+    sys.path.append(
+            os.path.join(os.path.expandvars('$HUNTSMAN_POCS'), 'scripts'))
```
AnthonyHorton: This is unnecessary, I think? You're not executing any scripts, just calling the main function.

scripts/pyro_camera_server.py
```diff
@@ -5,11 +5,19 @@
...
+    from auto_sshfs_mount import MountSSHFS
```
AnthonyHorton: imports should go with the others, at the top of the file.

scripts/pyro_camera_server.py
```diff
@@ -5,11 +5,19 @@
...
+    MountSSHFS()
```
AnthonyHorton: This probably should go in the `run_camera_server` function instead of this script: https://github.com/AstroHuntsman/huntsman-pocs/blob/develop/huntsman/utils/pyro/__init__.py#L114

That way the automounting will still happen if the camera server is started by calling the function directly instead of this script, and you can add attempted unmounting to the clean up code that gets executed when the camera server is stopped (here: https://github.com/AstroHuntsman/huntsman-pocs/blob/develop/huntsman/utils/pyro/__init__.py#L150)

AnthonyHorton (changes requested):
Some changes needed to keep in line with conventions used in POCS & huntsman-pocs. Also tests are failing.

AnthonyHorton:
@danjampro Well, first reason the Travis tests are failing is that we're not installing sshfs in the Travis virtual machine. Can fix that by adding `sudo apt-get install sshfs` to the `.travis.yml` file. 

Edit: The `.travis.yml` file from POCS shows how to add apt installed dependencies: https://github.com/panoptes/POCS/blob/develop/.travis.yml

lspitler:
Thanks for the detailed review @AnthonyHorton . @danjampro let us know if you have any questions about where things should go, etc.

scripts/auto_sshfs_mount.py
danjampro: We could do this later. But at the moment I think its more flexible to have the whole filesystem mounted in case we end up using it for something other than synching images.

scripts/pyro_camera_server.py
danjampro: /scripts is not installed as a python module, therefore I would not expect it to be on the default path.

huntsman/utils/pyro/__init__.py
```diff
@@ -151,3 +154,8 @@ def run_camera_server(ignore_local):
             print('\nShutting down...')
             name_server.remove(name=config['name'])
             print('Unregistered from name server')
+
+    #Unmount the SSHFS?
+    if unmount_sshfs:
+        sshfs_mount.unmount(mountpoint)
+
```
AnthonyHorton: This should go inside the `finally` block above so that it always gets executed when the daemon exits, regardless of reason.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+    logger('Done mounting SSHFS!')
+    
+    return mountpoint
+    
+#==============================================================================
+#==============================================================================
+
+if __name__ == '__main__':
+    
+    mount_sshfs()
```
AnthonyHorton: This bit _does_ belong in the scripts directory. What I meant was splitting this into two files. One in utils, containing the functions, and one in scripts that has the shebang, imports, `if __name__ == '__main__':`, and function call.

This follows the pattern established by @wtgee in POCS, and also followed by the Pyro name server and camera server stuff in huntsman-pocs.  Utility functions go in utils, scripts should go in scripts, and scripts should contain only the essentials, importing the functions they need from utils.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+            pass
+        
+#==============================================================================
+
+def mount_sshfs(logger=None, user=None):
+    '''
+    
+    '''
+    if logger is None:
+        logger = print
```
AnthonyHorton: All this logging stuff is going to fail if you do pass in a `Logger` object, because it's not a callable itself. You have to call the methods, like `.debug()`, `.error()`, etc.  If think the simplest thing to do is to put the print statements back and just have a conditional logging command for each one, e.g.
```
msg = f'Attempting to unmount {mountpoint}...'
print(msg)
if logger:
    logger.debug(msg)
```
It's clunky, but it would work.

The elegant solution would be to make a DummyLogger class, with `.debug()`, `.info()`, `warning()`, and so on, methods that just print the argument. Then you make the default logger for this function be an instance of this class.  That would save lots of ugly, duplicated code cluttering up this function but if you're going to put that much effort in you might as well just connect to the pyro-camera-server logger instead.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+
+    #Get SSH username from environment?
+    if user is None:
+          user = os.environ['PANUSER']
```
AnthonyHorton: Might want to throw an exception here if user has not been specified and the environment variable hasn't been set either.  A `warnings.warn()` at the very least.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+
+def mount(mountpoint, remote, server_alive_interval=20,
+          server_alive_count_max=3, logger=None):
+    '''
+    Mount remote on local.
+    '''
+    if logger is None:
+        logger = print
+        
+    if not os.path.isdir(mountpoint):
```
AnthonyHorton: This is redundant, because `os.makedirs` also checks whether mountpoint already exists.

In general the practice of checking if you can do something, then doing it, is best avoided because it leaves you open to race conditions. Best to just do the thing, and if necessary catch exceptions.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            os.makedirs(mountpoint, exist_ok=True)
+        except FileExistsError:
+            pass #For some reason this is necessary
```
AnthonyHorton: That's very weird, because this works just fine: https://github.com/panoptes/POCS/blob/develop/pocs/utils/images/fits.py#L360

Really you shouldn't need either the `try` or the `if`, this should all work with the single line `os.makedirs(mountpoint, exist_ok=True)`.

Should probably be setting the mode, though. It defaults to 777.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+def unmount(mountpoint):
+    '''
+    Unmount remote from local.
+    '''
+    if os.path.isdir(mountpoint):
+        options = ['fusermount', '-u', mountpoint]
+        try:
+            subprocess.run(options, shell=False, check=True)
+        except:
+            pass
```
AnthonyHorton: Maybe merits a warning if the mountpoint exists but can't be unmounted?

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+        except FileExistsError:
+            pass #For some reason this is necessary
+        
+    options = f'ServerAliveInterval={server_alive_interval},' + \
+              f'ServerAliveCountMax={server_alive_count_max}'
+    options = ['sshfs', remote, mountpoint, '-o', options]
+    try:
+        subprocess.run(options, shell=False, check=True)
+    except Exception as e:
+        logger(f'Failed to mount {remote} on {mountpoint}.')
```
AnthonyHorton: Include the exception in the message. Kind of moot if you're just printing it, but if you were reviewing a log you'd want to see the full exception here too.

huntsman/utils/sshfs_mount.py
```diff
@@ -0,0 +1,109 @@
...
+    if logger is None:
+        logger = print
+
+    #Get SSH username from environment?
+    if user is None:
+          user = os.environ['PANUSER']
+        
+    #Specify the mount point as ~/Huntsmans-Pro 
+    home = os.path.expandvars('$HOME')
+    mountpoint = os.path.join(home, 'Huntsmans-Pro')
```
AnthonyHorton: Probably should be getting the mountpoint from the config too, or at least have the option to override the default using the config.

scripts/pyro_camera_server.py
```diff
@@ -6,10 +6,11 @@
 but before POCS.
 """
 import argparse
-
 from huntsman.utils.pyro import run_camera_server
 
 if __name__ == "__main__":
+        
+    #Run the camera server
```
AnthonyHorton: Eventually... the code immediately following this comment is command line argument parsing, though.

AnthonyHorton (commented):
Looking good. I've suggested a few more, mostly minor changes, but I think this is almost good to go.

AnthonyHorton (approved):

danjampro:
Merging following approval.

PR #136, opened 2019-12-18
moved ip to camera conf

danjampro:
Removes the need for a huntsman_local.yaml. huntsman_pro_ip is now defined in pyro_camera_local.yaml, which needs to be updated on the pis.


lspitler (approved):
LGTM

PR #139, opened 2019-12-20
Config client

danjampro:
Camera server now retrieves the config file from the config server using its own IP via the panoptes.utils module (the config server still needs to be implemented - Lee was working on it).

### DIFF
--- a/huntsman/utils/__init__.py
+++ b/huntsman/utils/__init__.py
@@ -13,7 +13,6 @@ def query_config_server(logger=None, *args, **kwargs):
     try:
         key = get_own_ip()
         config = client.get_config(key, *args, **kwargs)
-        return config
     except Exception as e:
         msg = f'Unable to retrieve config file from server: {e}'
         if logger is not None:
@@ -21,6 +20,7 @@ def query_config_server(logger=None, *args, **kwargs):
         else:
             print(msg)
         raise(e)
+    return config
         
         
 def load_config(config_files=None, **kwargs):
--- a/huntsman/utils/__init__.py
+++ b/huntsman/utils/__init__.py
@@ -1,28 +1,9 @@
 import os
 from pocs.utils import listify
 from pocs.utils.config import load_config as config_loader
-from panoptes.utils.config import client
-from huntsman.utils.pyro import get_own_ip
 
 #==============================================================================
 
-def query_config_server(logger=None, *args, **kwargs):
-    '''
-    Query the config server using the IP of the client as the key.
-    '''
-    try:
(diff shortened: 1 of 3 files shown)


huntsman/utils/__init__.py
```diff
@@ -1,17 +1,41 @@
...
+from huntsman.utils.pyro import get_own_ip
 
+#==============================================================================
 
+def query_config_server(logger=None, *args, **kwargs):
+    '''
+    Query the config server using the IP of the client as the key.
+    '''
+    try:
+        key = get_own_ip()
```
lspitler: Move this out of the try area as if it errors, the error message is specific to `client` failing.

huntsman/utils/pyro/__init__.py
```diff
@@ -128,7 +128,10 @@ def run_camera_server(ignore_local, unmount_sshfs=True):
     mountpoint = sshfs_mount.mount_sshfs()
     
     Pyro4.config.SERVERTYPE = "multiplex"
-    config = load_config(config_files=['pyro_camera.yaml'], ignore_local=ignore_local)
```
lspitler, line 131: it might be handy to keep the functionality of loading a specific yaml file instead. Indeed, if you merge this now the system wouldn't work at all.

Let's default to using the local file for now (so `use_server=False`), until we get the config_server online.

huntsman/utils/__init__.py
```diff
@@ -1,17 +1,41 @@
 import os
 from pocs.utils import listify
 from pocs.utils.config import load_config as config_loader
+from panoptes.utils.config import client
+from huntsman.utils.pyro import get_own_ip
 
+#==============================================================================
 
+def query_config_server(logger=None, *args, **kwargs):
```
lspitler: Actually this entire function is redundant with: https://github.com/panoptes/panoptes-utils/blob/develop/panoptes/utils/config/client.py#L6
which already catches errors.

I think you just need `config = client.get_config(ip_address)` (note I removed kwargs as `get_config` doesn't accept it.

huntsman/utils/pyro/__init__.py
```diff
@@ -128,7 +128,10 @@ def run_camera_server(ignore_local, unmount_sshfs=True):
...
+        
+    #Load the config file from remote server
+    config = query_config_server()
```
lspitler:
```suggestion
    config = client.get_config(ip_address)
```

lspitler (approved):
A few suggestions to address, including one to remove `query_config_server` entirely. Can we test this before merging too?

PR #141, opened 2020-01-02
Config server

danjampro:
-Moved get_own_ip to huntsman.utils.__init__

- Created huntsman.utils.config to house the new pyro config server

- Updated run_camera_server to use the pyro config server

- Added a script to run the config server (/scripts/pyro_config_server.py)

I tested this on one of the pis and successfully retrieved a config.


huntsman/utils/config.py
```diff
@@ -0,0 +1,135 @@
...
+    '''
+    Class representing the config server.
+    '''
+    def __init__(self, config_file=None, **kwargs):
+        '''
+        
+        '''
+        if config_file is None:
+            config_file = os.path.join(os.environ['HUNTSMAN_POCS'],
+                                       'conf_files', 'camera_info_local.yaml')
```
lspitler: Change to `device_info_local.yaml` as per discussion on Slack.

Also, doesn't `load_config` search for both `device_info_local.yaml` and `device_info.yaml` and will default to `device_info_local.yaml` if it finds it? Shouldn't this like point to `device_info.yaml` instead of `device_info_local.yaml`?

huntsman/utils/config.py
```diff
@@ -0,0 +1,135 @@
...
+        '''
+        if key is None:
+            return self.config
+        return self.config[key]
+    
+#==============================================================================
+
+def locate_name_server(wait=None, logger=None):
+    '''
+    Locate and return the name server, waiting if necessary.
```
lspitler:
```suggestion
    Locate and return the name server (NS), waiting if necessary.
```

huntsman/utils/config.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            return Pyro4.locateNS()
+        except Pyro4.errors.PyroError:
+            msg = 'Unable to locate name server. Waiting...'
+            if logger is None:
+                print(msg)
+            else:
+                logger.debug(msg)
+            time.sleep(wait)
+
```
lspitler, line 80: Do you want to catch a `KeyboardInterrupt` so someone can cmd+c cancel this cleanly? Maybe returning a None so the routine calling it knows it should shut down?

huntsman/utils/config.py
```diff
@@ -0,0 +1,135 @@
...
+        except Pyro4.errors.PyroError:
+            msg = 'Unable to locate name server. Waiting...'
+            if logger is None:
+                print(msg)
+            else:
+                logger.debug(msg)
+            time.sleep(wait)
+        
+        
+def start_config_server(host=None, port=6563, name='config_server',
```
lspitler, line 82: should this and the next function be a method of the `ConfigServer` class?

huntsman/utils/pyro/__init__.py
```diff
@@ -1,77 +1,13 @@
...
 from Pyro4 import naming, errors
 
-from huntsman.utils import load_config, sshfs_mount
+from huntsman.utils import get_own_ip, load_config, sshfs_mount
+from huntsman.utils.config import query_config_server
 from huntsman.camera.pyro import CameraServer
-from panoptes.utils.config import client
-
-
-def get_own_ip(verbose=False, logger=None):
```
lspitler, line 12: If you haven't already, can you make sure that . any other use of `get_own_ip` now points at the new location?

scripts/pyro_config_server.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+"""
+Script to start the config server. Make sure the Pyro name server is already
```
lspitler: If it starts it, then how about renaming to `start_pyro_config_server.py` or even just `start_config_server.py`

scripts/pyro_config_server.py
```diff
@@ -0,0 +1,16 @@
...
+now.
+"""
+from huntsman.utils.config import start_config_server
+
+#==============================================================================
+
+if __name__ == "__main__":
+            
+    #Start the config server
+    start_config_server()
```
lspitler: Need EOL at EOF

huntsman/utils/pyro/__init__.py
```diff
@@ -173,4 +109,9 @@ def run_camera_server(ignore_local, unmount_sshfs=True, use_server=False):
             #Unmount the SSHFS?
             if unmount_sshfs:
                 sshfs_mount.unmount(mountpoint)
+                    
+#==============================================================================
+       
+        
+
```
lspitler, line 117: Need EOL at EOF

lspitler (approved):
Looks good @danjampro . A few suggestions before you merge.

huntsman/utils/config.py
danjampro: Nice idea - yes, probably. The only thing that stopped me doing this is the need to register a ConfigServer object with the NS. Perhaps self could be used, but this is probably simpler for now

huntsman/utils/config.py
danjampro: Sounds good, but at the moment I don't think there would be much difference. I also think returning None rather than raising an error might cause more confusion down the line. Do you have a specific scenario where this would be useful in-mind?

lspitler: Agree returning None is arbitrary and could cause issues @danjampro . How cntrl+C triggers an `sys.exit(0)` call with a kind message saying shutting down name server client.

huntsman/utils/config.py
lspitler: I see. Makes sense.

huntsman/utils/config.py
lspitler: Oh and make sure to send the shutdown message to the logger as well @danjampro

PR #142, opened 2020-01-07
Integrate config server with camera server

danjampro:
Camera server and SSHFS now use the config server to get their config by default. This behaviour can be overridden by specifying the config_files keyword argument.

I have tested this on the control computer and huntsman pi 000 and I was able to properly start the camera server.

I'm not sure it will pass tests as there is not yet support in Travis for the config server...


huntsman/utils/config.py
```diff
@@ -144,3 +146,47 @@ def query_config_server(key=None, name='config_server'):
...
+    '''
+    if logger is None:
+        logger = DummyLogger()
+        
+    #Load config from local files?
+    if config_files is not None:
+        logger.debug(f'Loading config from local file(s).')
+        try:
+            config = load_config(config_files, **kwargs)
+        except Exception as e:
```
lspitler: As this really just calls https://github.com/panoptes/POCS/blob/develop/pocs/utils/config.py#L11 which has extensive error handling, I don't think you need a try/except here as its redundant

huntsman/utils/config.py
```diff
@@ -144,3 +146,47 @@ def query_config_server(key=None, name='config_server'):
...
+            logger.error(f'Unable to load local config file(s): {e}')
+            raise(e)
+    
+    #Load config from the config server?
+    else:
+        if key is None:
+            key = get_own_ip()
+        logger.debug(f'Loading remote config with key: {key}')
+        try:
```
lspitler: ditto - this might be redundant if `query_config_server` already has appropriate error handling, as it should.

huntsman/utils/config.py
```diff
@@ -144,3 +146,47 @@ def query_config_server(key=None, name='config_server'):
...
+        The config dictionary.
+    '''
+    if logger is None:
+        logger = DummyLogger()
+        
+    #Load config from local files?
+    if config_files is not None:
+        logger.debug(f'Loading config from local file(s).')
+        try:
+            config = load_config(config_files, **kwargs)
```
lspitler: Don't you need to query by ip here as well? Or at least do a:
```suggestion
            config = load_config(config_files, **kwargs)[key]
```

huntsman/utils/config.py
```diff
@@ -144,3 +146,47 @@ def query_config_server(key=None, name='config_server'):
...
+        except Exception as e:
+            logger.error(f'Unable to load local config file(s): {e}')
+            raise(e)
+    
+    #Load config from the config server?
+    else:
+        if key is None:
```
lspitler: move this out of if/else if you need to check this in both the if/else cases

huntsman/utils/config.py
```diff
@@ -144,3 +146,47 @@ def query_config_server(key=None, name='config_server'):
...
+            key = get_own_ip()
+        logger.debug(f'Loading remote config with key: {key}')
+        try:
+            config = query_config_server(key=key)
+        except Exception as e:
+            logger.error(f'Unable to load remote config: {e}')
+            raise(e)
+            
+    return config
```
lspitler, line 197: maybe simplify the above with:
```suggestion
    return config[key]
```

huntsman/utils/config.py
```diff
@@ -110,15 +113,14 @@ def start_config_server(host=None, port=6563, name='config_server',
         #Register with pyro & the name server
         uri = daemon.register(config_server)
```
lspitler, line 121: Consider using a more obvious variable than `uri` unless this is standard in pyro.

huntsman/utils/pyro/__init__.py
```diff
@@ -89,22 +90,23 @@ def run_camera_server(ignore_local, unmount_sshfs=True, use_server=False):
...
         try:
             daemon.requestLoop()
         finally:
-            print('\nShutting down...')
+            logger.info('\nShutting down...')
             name_server.remove(name=config['name'])
-            print('Unregistered from name server')
+            logger.info('Unregistered from name server')
 
             #Unmount the SSHFS?
```
lspitler, line 109: I see a '?'. Just making sure this is OK

lspitler (approved):
@danjampro looks good. Make sure Travis works before merging...

PR #145, opened 2020-01-12
Generalised device startup with docker

danjampro:
The idea of this PR is that each device (e.g. camera server, dome server) should contain only minimal information in order to download & run the most up-to-date version of a common docker image. This is encapsulated in scripts/run_device_container.sh.

The code run in the docker container itself is responsible for telling the device what it is and what command(s) it should execute. This is the purpose of scripts/startup_device.py. 

The docker GitHub workflow has been updated to reflect these changes.


docker/device_startup.docker
```diff
@@ -79,5 +79,5 @@ RUN pip install pyyaml
 #For some reason this is necessary
 RUN rm -r $PANDIR/images
 
-#Mount the SSHFS & start pyro camera server (blocking)
-CMD python $HUNTSMAN_POCS/scripts/pyro_camera_server.py 
+#Startup the device
+CMD python $HUNTSMAN_POCS/scripts/startup_device.py
```
wtgee:
```suggestion
CMD ["python", "$HUNTSMAN_POCS/scripts/startup_device.py"]
```

See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#cmd

.github/workflows/publish_device_startup.yml
```diff
@@ -24,8 +24,8 @@ jobs:
           docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
           docker buildx build \
             --platform linux/arm64 \
-            --tag huntsmanarray/camera_server:latest \
+            --tag huntsmanarray/device_startup:latest \
```
wtgee, line 27: `huntsmanarray/device`? This does more than just start each device, it also keep devices running.

scripts/run_device_container.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+#This script should be all any device (other than the control computer)
+#needs to run. The Pyro name server and config server should ideally be
+#initialised beforehand (e.g. on the control computer).
+
+#Download the latest docker image from docker hub
+docker pull huntsmanarray/device_startup
```
wtgee:
```suggestion
docker pull huntsmanarray/device_startup:latest
```

scripts/run_device_container.sh
```diff
@@ -0,0 +1,13 @@
...
+
+#Run the docker container
+#Note that network=host to map container's ports to the hosts directly
+#This is OK for the pis etc. but probably not for the control computer
+docker run huntsmanarray/device_startup:latest --network host
```
wtgee:
```suggestion
docker run huntsmanarray/device_startup:latest --network host --privileged -it
```
Is this working without?  `privileged` will allow you to access the device nodes, which it would seem like we need. `it` allows for `interactive` and `terminal` (or something, that's how I remember it).

wtgee (approved):
Might be worth starting to check out [`docker-compose`](https://docs.docker.com/compose/). No specific usage I see yet, but it's more common to use even for simple situations and might help as this gets more complex.

scripts/run_device_container.sh
danjampro: I think I had it working previously without --privileged and I have read that its best to avoid if possible. So, let's test it without to begin with.

PR #146, opened 2020-01-20
Automatic device startup

danjampro:
Make run_device_container.sh run startup_device.py without human intervention.

- Various changes to the dockerfile

- Added ability to wait for config server if it is not initially found

- Replaced prints with DummyLogger

- Disabled ssh strict host key checking from inside the container

- Enabled container to use identity of host for SSH connections


docker/device_startup.docker
```diff
@@ -1,54 +1,50 @@
...
-RUN ./Berryconda3-2.0.0-Linux-armv7l.sh -b -p $CONDA_PATH
+ENV CONDA_PATH=/$PANDIR/berryconda3
+RUN wget $CONDA_URL && chmod +x Berryconda3-2.0.0-Linux-armv7l.sh && \
+    ./Berryconda3-2.0.0-Linux-armv7l.sh -b -p $CONDA_PATH
 
-RUN ln -s $CONDA_PATH/bin/conda /usr/local/bin/
-RUN ln -s $CONDA_PATH/bin/python3.6 /usr/local/bin/python
-RUN ln -s $CONDA_PATH/bin/pip /usr/local/bin/
+USER root
+RUN ln -s $CONDA_PATH/bin/conda /usr/local/bin/ && \
```
lspitler, line 32: I think `conda activate` does this linking for you. https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment

docker/device_startup.docker
```diff
@@ -1,54 +1,50 @@
...
-ENV PANDIR=/var/huntsman
 ENV POCS=${PANDIR}/POCS
 ENV PAWS=${PANDIR}/PAWS
 ENV HUNTSMAN_POCS=${PANDIR}/huntsman-pocs
 
 #POCs folders
-RUN mkdir $PANDIR && chown $PANUSER $PANDIR
-RUN mkdir -p $PANDIR/logs
-RUN mkdir -p $PANDIR/images
+RUN mkdir -p $PANDIR/logs && mkdir -p $PANDIR/images
```
lspitler, line 47: what happens to the logs in `$PANDIR/logs` after this docker instance is shutdown? Are the logs lost forever? Maybe we should make them a `sshfs` access point like you do for the `$PANDIR/images` directory?

docker/device_startup.docker
```diff
@@ -64,20 +60,21 @@ RUN python setup.py develop
...
-RUN pip install pyyaml
+RUN chmod 1777 /tmp 
+RUN chown -R $PANUSER $PANDIR && mkdir /home/$PANUSER \
+    && chown -R $PANUSER /home/$PANUSER
+USER $PANUSER
 
 #==============================================================================
 
 #For some reason this is necessary
 RUN rm -r $PANDIR/images
```
lspitler, line 77: you just created this... is this because the soft link(?) to `sshfs` mount point can't be created later on?

huntsman/utils/sshfs_mount.py
```diff
@@ -26,9 +32,13 @@ def mount(mountpoint, remote, server_alive_interval=20,
...
+    #SSH options
+    strict_host_key_checking = "yes" if strict_host_key_checking else "no"
     options = f'ServerAliveInterval={server_alive_interval},' + \
-              f'ServerAliveCountMax={server_alive_count_max}'
+              f'ServerAliveCountMax={server_alive_count_max},' + \
+              f'StrictHostKeyChecking={strict_host_key_checking}'
     options = ['sshfs', remote, mountpoint, '-o', options]
+    
     try:
         subprocess.run(options, shell=False, check=True)
```
lspitler, line 43: What happens if the ssh link falls over? Will entire docker image crash or will it wait until it reappears?

scripts/run_device_container.sh
```diff
@@ -8,6 +8,43 @@
...
 #Note that network=host to map container's ports to the hosts directly
-#This is OK for the pis etc. but probably not for the control computer
-docker run huntsmanarray/device_startup:latest --network host -it
+#This is OK for the pis etc. but probably not for the control computer.
+
+#-it allows the container to run in interactive mode.
+
+#--cap-add SYS_ADMIN --device /dev/fuse is necessary for the SSHFS mount.
+
+#-v ~/.ssh:/root/.ssh allows the container to use the credentials of the host
```
lspitler, line 19: too much commented out stuff here. If these are notes to self for control computer docker work, maybe put the comments into that new file?

lspitler (approved):
LGTM @danjampro . A few suggestions for you to consider.

docker/device_startup.docker
wtgee: The way that I have been doing things is to map the host `$PANDIR` to the `$PANDIR` in the container. This is actually how you use it as a development mode as well. So really all of `$PANDIR` comes from the host and all of the system dependencies exist in the docker image.

docker/device_startup.docker
wtgee: Uh, yeah, what's the story here?

huntsman/utils/config.py
```diff
@@ -151,17 +148,30 @@ def query_config_server(key=None, name='config_server', logger=None):
...
+        
+        try:
+            config_server = Pyro4.Proxy(f'PYRONAME:{name}')
+            return config_server.get_config(key=key)
+        
+        except Pyro4.errors.NamingError as e:
+            
+            if wait is not None:
+                logger.info('Failed to locate config server. \
+                            Waiting %is before retrying.'%wait)
```
wtgee: It's fine to use the logger style formatting but in general it's better to be consistent and just use f-strings or `.format`. It's true that there are some slight advantages to letting the logger do it but it's not as robust.

scripts/run_device_container.sh
```diff
@@ -8,6 +8,43 @@
...
+CONTAINER_ID=$(docker create -it --network host --cap-add SYS_ADMIN \
+        --device /dev/fuse huntsmanarray/device_startup:latest /bin/bash)
+
+#Start the container
+docker start $CONTAINER_ID
+
+#Extract some environment variables from the container
+CONTAINER_USER=$(docker exec $CONTAINER_ID echo $USER)
+CONTAINER_HOME=$(docker exec $CONTAINER_ID echo $HOME)
+CONTAINER_PANDIR=$(docker exec $CONTAINER_ID echo $PANDIR)
```
wtgee: I'm not sure this is doing the right thing. Those variables after echo will probably be expanded before-hand and they all exist on the host. Have you double-checked this?

scripts/run_device_container.sh
```diff
@@ -8,6 +8,43 @@
...
+docker cp ~/.ssh $CONTAINER_ID:$CONTAINER_HOME/.ssh
+
+#Allow ssh credentials to be accessed by huntsman user
+docker exec -u root $CONTAINER_ID chown -R $CONTAINER_USER $CONTAINER_HOME/.ssh
+
+#Run the device script
+docker exec -it $CONTAINER_ID python $CONTAINER_PANDIR/huntsman-pocs/scripts/startup_device.py
+
+#Finally, kill the container
+docker exec -u root $CONTAINER_ID pkill bash
```
wtgee, line 50: We should try to use [`docker-compose`](https://docs.docker.com/compose/) for most of what this file does. It's more standard and more flexible. This doesn't apply to the last line but really the whole file.

Is there something complicated this script is trying to achieve that I don't see or is it just starting the container?

wtgee (approved):
The Dockerfile still really needs to be cleaned up to minimize the `RUN` statements, but that can be left for a different PR.

I made the note about the script that runs the images, which seems a bit unnecessary, or at least complicated.

docker/device_startup.docker
lspitler: This makes sense. It also might solve a question for how to link up the to-be-developed-soon control computer docker instance to the yaml files created on-the-fly from @Physarah 's EMS: https://github.com/AstroHuntsman/huntsman-ems

huntsman/utils/config.py
danjampro: Interesting what you say about PANDIR, although I don't quite understand. Where does e.g. Huntsman-POCS get installed in the container if PANDIR is (presumably) a mounted volume? I will take a look at the PANOPTES stuff. 

As @lspitler mentioned, there was some weird error while mounting the SSHFS and this was a temporary work around. Will take a look, but perhaps this can be part of the next PR in which the docker stuff is cleaned up a little (perhaps including docker-compose, which I need to read up on).

scripts/run_device_container.sh
danjampro: Excellent point and very well noticed - I think you're right. I'll change it before merging.

scripts/run_device_container.sh
danjampro: The only reason I'm using exec is that, in order to share the ssh keys, it is required to know where to put them and which user requires them before hand. So this is a way of doing it without the script needing to assume environment variables inside the container. Will take a look at docker-compose.

docker/device_startup.docker
danjampro: Will take a look. It does seem a bit unnecessary.

docker/device_startup.docker
danjampro: @lspitler I imagined those files to be distributed using the config server?

huntsman/utils/sshfs_mount.py
danjampro: It is configured to retry the connection periodically, although this expires after a few minutes in which case it will cause problems. It can probably (easily) be made a bit more robust in future.

PR #150, opened 2020-02-02
NGAS IP lookup in Camera server

danjampro:
With the new config server update, the NGAS ip wasn't being found by the cameras and the NGAS push failed. Now the config server is responsible for looking up the NGAS IP remotely and passing the information to the Camera object when it is created via its config dictionary.


lspitler (approved):
LGTM

PR #151, opened 2020-02-04
Install ZWO camera library

danjampro:
I have added the ZWO library to the docker image in /usr/local/lib as per the instructions in [POCS](https://github.com/panoptes/POCS/wiki/Additional-Hardware-Drivers-Installation). The updated docker image successfully builds on my laptop.

I originally tested the installation on Huntsmanpi005 and the autoguider ZWO and it worked. 

Also added some storage saving measures by reducing the number of layers.


lspitler (approved):
LGTM @danjampro . Is it easy to add FLI library build now? Or is that another PR?

PR #152, opened 2020-02-04
Install ZWO filterwheel library

danjampro:
Added filter wheel install into device docker image.

Docker build has been successfully tested on my laptop.

This needs testing with the actual hardware, but it might be easiest to do this after merging.


danjampro:
Merging following @AnthonyHorton's  approval.

PR #154, opened 2020-02-05
Allow docker container to access host devices

danjampro:
--privileged allows the container to use devices of host.

Also added pyserial module required by POCS focuser module.

### DIFF
--- a/docker/device_startup.docker
+++ b/docker/device_startup.docker
@@ -39,7 +39,7 @@ RUN ln -s $CONDA_PATH/bin/conda /usr/local/bin/ && \
 RUN pip install --upgrade pip && \
     pip install setuptools --upgrade && \
     conda install astropy matplotlib requests scipy && \
-    pip install pymongo netifaces pyro4 pyyaml
+    pip install pymongo netifaces pyro4 pyyaml pyserial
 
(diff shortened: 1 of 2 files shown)


AnthonyHorton (approved):
LGTM, once the tests pass.

danjampro:
> LGTM, once the tests pass.

@AnthonyHorton Do we need to wait for tests to pass? This is not covered by Travis, the real test is whether the docker image builds (see the actions tab), which it does. So I think this can be merged.

AnthonyHorton:
> > LGTM, once the tests pass.
> 
> @AnthonyHorton Do we need to wait for tests to pass? This is not covered by Travis, the real test is whether the docker image builds (see the actions tab), which it does. So I think this can be merged.

Well, this PR was to allow the Docker image to access devices on the host. Does the Docker image built test that? Also the actions tab is telling me that the latest Docker image build failed?

The unit tests are important, that's how we avoid accidentally breaking stuff. I know we're in a hurry right now, and it wasn't this PR that broke them, but it does bother me that the tests broke in `develop` but we're just merging more PRs anyway with no way to tell if we're breaking more things in the process.

PR #158, opened 2020-02-07
Clean up of Pyro code

AnthonyHorton:
Clean up of Pyro camera and focuser code, and associated tests, prior to adding filterwheel support.

### DIFF
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -322,44 +322,6 @@ class Camera(AbstractCamera):
         assert self.focuser.is_connected, \
             self.logger.error("Focuser must be connected for autofocus.")
 
-        # Make certain that all the argument are builtin types for easy Pyro serialisation
-        if isinstance(seconds, u.Quantity):
-            seconds = seconds.to(u.second).value
-        if seconds is not None:
-            seconds = float(seconds)
-
-        if focus_range is not None:
-            focus_range = (int(limit) for limit in focus_range)
-
-        if focus_step is not None:
-            focus_step = (int(step) for step in focus_step)
-
-        if keep_files is not None:
-            keep_files = bool(keep_files)
-
-        if take_dark is not None:
-            take_dark = bool(take_dark)
-
-        if thumbnail_size is not None:
-            thumbnail_size = int(thumbnail_size)
-
-        merit_function = str(merit_function)
-        merit_function_kwargs = dict(merit_function_kwargs)
-
-        if mask_dilations is not None:
-            mask_dilations = int(mask_dilations)
-
-        if coarse is not None:
-            coarse = bool(coarse)
-
-        if make_plots is not None:
-            make_plots = bool(make_plots)
-
-        if isinstance(timeout, u.Quantity):
-            timeout = timeout.to(u.second).value
-        if timeout is not None:
-            timeout = float(timeout)
-
         # Compile aruments into a dictionary
         autofocus_kwargs = {'seconds': seconds,
                             'focus_range': focus_range,
@@ -373,8 +335,6 @@ class Camera(AbstractCamera):
                             'coarse': coarse,
                             'make_plots': make_plots}
         autofocus_kwargs.update(kwargs)
-
-        focus_dir = os.path.join(os.path.abspath(self.config['directories']['images']), 'focus/')
 
         # Make sure proxy is in async mode
         Pyro4.asyncproxy(self._proxy, asynchronous=True)
--- a/huntsman/camera/pyro.py
+++ b/huntsman/camera/pyro.py
@@ -63,17 +63,11 @@ class Camera(AbstractCamera):
 
     @property
     def egain(self):
-        if self._egain is not None:
-            return self._egain
-        else:
-            raise NotImplementedError
+        self._proxy.egain
 
     @property
     def bit_depth(self):
-        if self._bit_depth is not None:
-            return self._bit_depth
-        else:
-            raise NotImplementedError
+        return self._proxy.bit_depth
 
     @property
     def temperature(self):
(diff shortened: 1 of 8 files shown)


AnthonyHorton:
In order to get the tests actually running again I've set two tests from `test_huntsman.py` to skip for now (`test_run_wait_until_safe` and `test_run`). Both of these hang indefinitely at the moment. We need to work out what it going wrong and fix them asap, but perhaps not in this PR.

danjampro:
Looks good @AnthonyHorton . I have only one main comment before I approve - and that is whether there is a missing `return` statement (see above). I think I will remove yet more lines in my pending PR which removes the _file_transfer and _cleanup_directories functions in `pyro.Camera`.

huntsman/camera/pyro.py
```diff
@@ -23,8 +24,24 @@
...
+            "yaml_dump": ayaml.dump(obj)}
+
+
+def dict_to_astropy(class_name, d):
+    """De-serialiser function for Astropy objects using astropy.io.misc.yaml.load()."""
+    return ayaml.load(d["yaml_dump"])
+
+SerializerBase.register_class_to_dict(u.Quantity, astropy_to_dict)
+SerializerBase.register_dict_to_class("astropy_yaml", dict_to_astropy)
+
```
wtgee, line 41: Is this custom to Pyro or just custom because of astropy quantities?  Seems like this doesn't really belong in the camera class itself.

huntsman/camera/pyro.py
```diff
@@ -327,44 +307,6 @@ def autofocus(self,
...
-            coarse = bool(coarse)
-
-        if make_plots is not None:
-            make_plots = bool(make_plots)
-
-        if isinstance(timeout, u.Quantity):
-            timeout = timeout.to(u.second).value
-        if timeout is not None:
-            timeout = float(timeout)
-
```
wtgee, line 367: Just curious as to why we are able to remove all this now if was needed earlier.

huntsman/tests/test_huntsman.py
```diff
@@ -249,10 +249,12 @@ def test_pyro_camera(config, camera_server):
     assert obs.cameras['camera.simulator.001'].is_connected
 
 
+@pytest.mark.skip
 def test_run_wait_until_safe(observatory, cmd_publisher, msg_subscriber):
     os.environ['POCSTIME'] = '2016-09-09 10:00:00'
 
-    observatory.db.insert_current('weather', {'safe': False})
+    # Make sure DB is clear for current weather
+    observatory.db.clear_current('weather')
```
wtgee: Did this change actually do anything or was it just the attempt to fix this test?

huntsman/tests/test_huntsman.py
```diff
@@ -288,7 +290,7 @@ def start_pocs():
         assert wait_for_running(msg_subscriber)
         observatory.logger.info('Got RUNNING message')
 
-        time.sleep(5)
+        time.sleep(2)
```
wtgee: What's reason for making this less?  If anything I would suspect the test is hanging because this sleep time is too short. 

In general though, if these changes to the test didn't fix the test and then we are skipping the test anyway, I would prefer not to change the existing code unless we know that actually accomplishes something.  In theory it was working at some point with the code as is, so I wouldn't want to complicate the debugging.

states/huntsman/sleeping.py
```diff
@@ -1,11 +1,15 @@
...
+                 "Stopping loop.")
         pocs.stop_states()
     else:
-        pocs.say("Another successful night!")
+        # Note: Unit will "sleep" before transition until it is safe
+        # to observe again.
+        pocs.next_state = 'ready'
+        pocs.reset_observing_run()
+
+    pocs.say("Another successful night!")
```
wtgee: This doesn't seem like it should be part of this PR.

wtgee (approved):
LGTM. I'm a bit confused as to _why_ you can clean up so much, especially the things that just seem to be safety type check of params, but the more that's removed while retaining functionality the better.

huntsman/camera/pyro.py
AnthonyHorton: Both? This registers custom serialiser & deserialiser functions with Pyro for use with astropy quantities. That's how I was able to remove all the repetitive code was just there to convert quantities to/from floats or ints, because now the remote method calls can use quantities as arguments or return values the same as the local equivalents.

The serialiser/deserialiser functions can be somewhere else (`pocs.utils.pyro` ?), but the registration is needed by both the `Camera` and `CameraServer` classes, which are both in this file.

Next on the todo list would be registering serialiser/deserialiser function for the POCS custom exceptions, so that we no longer get Pyro serialiser errors when a remote camera raises PanError, NoFound, etc.

huntsman/camera/pyro.py
AnthonyHorton: The quantities stuff is now redundant because of the quantity (de)serialiser. The other input validation isn't needed because the remote method call will raise exceptions for these, which will be re-raised locally. I think these were originally added either because of an over abundance of caution, or because I couldn't get the remote exception hook working properly. Either way, not needed now.

huntsman/tests/test_huntsman.py
AnthonyHorton: Just an attempt to fix the test. The version of this test in POCS does this, so I copied it to `huntsman-pocs` to see if it made a difference. It did, but the difference didn't fix the test.

huntsman/tests/test_huntsman.py
AnthonyHorton: Again, just copying something from POCS in a desperate attempt to fix the test.  I agree though, since these didn't work I should take these back out.

states/huntsman/sleeping.py
AnthonyHorton: Nope. More failed attempt to fix the tests, based on moving closer to POCS.

AnthonyHorton:
> LGTM. I'm a bit confused as to _why_ you can clean up so much, especially the things that just seem to be safety type check of params, but the more that's removed while retaining functionality the better.

Mostly it's a question of trusting the remote objects to do the input validation. That and removing the need to convert quantities to/from Python built in types for Pyro (de)serialisation.  The Pyro stuff can and should be a _very_ thin wrapper (in fact it can be even more minimal than that, but not if you need it to work with existing classes with no modifications as we did).

PR #160, opened 2020-02-10
SSHFS overhaul 

danjampro:
@AnthonyHorton noticed that re-running tests with the current implementation required manual removal of the link between the mounted images directory and the images directory. This PR addresses that issue by removing the linking process altogether and instead mounts the images directory on the control computer directly to the local images directory. These directories are defined in `device_info.yaml`. 

I have added the `control` entry to `device_info.yaml` which is where the control computer config should live. It might be good for the control computer to read its directories from this to avoid hard-code duplication of directory definitions for the control computer.

In order to facilitate proper reading of the nested configs in `device_info.yaml`, `pocs.utils.config._parse_config` now runs after calling `huntsman.config.query_config_server` and a `key` is specified.

Lastly, `huntsman_pro_ip` has been renamed to `control_ip` to reflect changes to the control computer, and this should be updated in the automatic config generation code @Physarah.

Deleting the redundant `_file_transfer` and `_cleanup_directories` calls were necessary to make tests pass, so I also deleted the function defs.


conf_files/device_info.yaml
```diff
@@ -28,5 +28,8 @@ localhost:
         base: /var/panoptes
         images: temp
         data: data
-messaging:
-    huntsman_pro_ip: localhost 
+control:
+    control_ip: localhost
```
wtgee: `control.control_ip` seems redundant. Why not `ip_address`?

huntsman/tests/test_camera.py
```diff
@@ -27,23 +27,25 @@
...
 ids = ['pyro']
 
-
+"""
 @pytest.fixture(scope='module')
 def images_dir(tmpdir_factory):
     directory = tmpdir_factory.mktemp('images')
     return str(directory)
-
+"""
```
wtgee: If this is not needed let's remove.

huntsman/utils/pyro/__init__.py
```diff
@@ -62,20 +62,20 @@ def run_camera_server(ignore_local=False, unmount_sshfs=True, logger=None,
...
     """
     if logger is None:
         logger = DummyLogger()
-        
-    #Mount the SSHFS
-    mountpoint = sshfs_mount.mount_sshfs()
-    
+
+    #Mount the SSHFS images directory
+    mountpoint = sshfs_mount.mount_images_dir()
```
wtgee: Looks like you should be passing the logger through to most of these.

```suggestion
    mountpoint = sshfs_mount.mount_images_dir(logger=logger)
```

wtgee (approved):
LGTM. You mention some changes to `_parse_config` order in the description but all I see here is it being removed completely.

danjampro:
> LGTM. You mention some changes to `_parse_config` order in the description but all I see here is it being removed completely.

Well spotted... seems I managed to accidentally revert those changes... They are back now!

wtgee:
@danjampro looks like some files are missing now? Lost in a rename or something.

danjampro:
> @danjampro looks like some files are missing now? Lost in a rename or something.

@wtgee I renamed `sshfs_mount.py` to `sshfs.py` and removed  `scripts/sshfs_mount.py` which is unused. Is that what you mean?

wtgee:
> > @danjampro looks like some files are missing now? Lost in a rename or something.
> 
> @wtgee I renamed `sshfs_mount.py` to `sshfs.py` and removed `scripts/sshfs_mount.py` which is unused. Is that what you mean?

Yeah, must be.

conf_files/device_info.yaml
```diff
@@ -28,5 +28,8 @@ localhost:
         base: /var/huntsman
         images: /tmp/images
         data: data
-messaging:
-    huntsman_pro_ip: localhost
+control:
+    ip_address: localhost
+    directories:
+        base: /var/huntsman
+        images: images
```
AnthonyHorton, line 35: Now that we can override the base images directory using the fixtures in conftest should probably have these two `images` directory setting be consistent. I changed to `/tmp/images` for the remote end in my PR so that the unit tests wouldn't write to the 'proper' images directory, but there is a better way of doing that now thanks to your changes to the config server and corresponding test fixture.

huntsman/camera/pyro.py
```diff
@@ -129,6 +129,9 @@ def is_exposing(self):
     def is_ready(self):
         '''
         True if camera is ready to start another exposure, otherwise False.
+
+        Check self.is_exposing first to side-step so-far unexplained
+        hanging possibly caused by is_temperature_stable.
```
AnthonyHorton: I don't think this comment should be in the docstring for this property. It strongly suggests to the user that they are responsible for checking `is_exposing` first before checking `is_ready`, but my previous PR built that check in to the `is_ready()` method (immediately below this docstring!) so that would now be redundant.

huntsman/tests/test_camera.py
```diff
@@ -30,14 +30,16 @@
 
 # Ugly hack to access id inside fixture
 @pytest.fixture(scope='module', params=zip(params, ids), ids=ids)
-def camera(request, images_dir, camera_server):
+def camera(request, images_dir, camera_server, config_server):
```
AnthonyHorton: Redundant because the `camera_server` fixture requires the `config_server` fixture now.

huntsman/tests/test_camera.py
```diff
@@ -30,14 +30,16 @@
...
     if request.param[0] == PyroCamera:
         ns = Pyro4.locateNS()
         cameras = ns.list(metadata_all={'POCS', 'Camera'})
         cam_name, cam_uri = cameras.popitem()
         camera = PyroCamera(port=cam_name, uri=cam_uri)
 
+    # Modify the client-side camera config
     camera.config['directories']['images'] = images_dir
```
AnthonyHorton: Shouldn't this be using the new `images_dir_local` fixture instead? This 'camera' is the Pyro proxy wrapper, the local interface for the distributed camera.

Actually, isn't this entirely unnecessary now that you're setting both the local and remote image directories in the config server fixture?

huntsman/utils/config.py
```diff
@@ -18,52 +19,64 @@ class ConfigServer():
...
         if config_file is None:
-            config_file = os.path.join(os.environ['HUNTSMAN_POCS'],
-                                       'conf_files', 'device_info.yaml')           
+            config_file = os.path.expandvars(os.path.join('$HUNTSMAN_POCS',
+                                            'conf_files', 'device_info.yaml'))
+
         #Read the config file
-        self.config_ = load_config(config_files=[config_file], **kwargs)
-        
+        self.config_ = load_config(config_files=[config_file], parse=self.parse,
```
AnthonyHorton: What's the story with the trailing underscore? Should it be a leading underscore instead, to make the config a pseudo-private member variable with access controlled via the property getter & setter?

AnthonyHorton (approved):
Looks good! A few small suggestions, but LGTM.

huntsman/utils/config.py
danjampro: Yep - will put it at the front.

conf_files/device_info.yaml
danjampro: @AnthonyHorton So are you suggesting to modify the `images` directory here? If so, I am a little confused because I thought the temp directories were generated by `pytest` and won't have the base/directory format?

AnthonyHorton: Suggesting changing the `images` entry on line 29 back from `/tmp/images` to `images`. I made that change because since the switch to sshfs the remote cameras couldn't use the temp directories generated by the test fixtures, and the only way I could make them use a temporary images directory was to change the default config file to point them to `/tmp`. Now that is no longer necessary thanks to your changes to the config server and the test fixtures we can change the default config file back to being an example of how a real config file would look, i.e. `base: /var/huntsman` and `images: images` for both ends of the distributed system.

AnthonyHorton:
I suspect several of these test failures (possibly including the autofocus one) are caused by the change in scope of the images directory tests fixtures to module. The previous scope made sure each camera test ran with it's own images directory so that there was no potential for filename collisions or confusion when checking for the presence of certain output files. With module scope all the tests in `test_camera.py` will send their output to the same temporary images directory. There's no reason that can't work, but it does require more careful cleanup and checking.

danjampro:
> I suspect several of these test failures (possibly including the autofocus one) are caused by the change in scope of the images directory tests fixtures to module. The previous scope made sure each camera test ran with it's own images directory so that there was no potential for filename collisions or confusion when checking for the presence of certain output files. With module scope all the tests in `test_camera.py` will send their output to the same temporary images directory. There's no reason that can't work, but it does require more careful cleanup and checking.

@AnthonyHorton Hmmm.... They were working alright before I merged with #158. 

I had to undo one of your changes to get `test_autofocus_coarse` working. Specifically, I removed the hardcoded images_dir directory in test_camera.patterns fixture. But now `test_autofocus_fine` hangs and for some reason I'm getting a key error in `test_observation`, which isn't making much sense to me at the moment...

AnthonyHorton:
> @AnthonyHorton Hmmm.... They were working alright before I merged with #158.
> 
> I had to undo one of your changes to get `test_autofocus_coarse` working. Specifically, I removed the hardcoded images_dir directory in test_camera.patterns fixture. But now `test_autofocus_fine` hangs and for some reason I'm getting a key error in `test_observation`, which isn't making much sense to me at the moment...

Yes, that hard coded images directory was the other half of my brute force method of making the remote cameras use `/tmp`, post sshfs. Your new way is better.

The hang I suspect is the result of a file name collision, but we'd have to run the test suite locally using your branch in order to see the logs from the remote camera to work out what is happening there. I don't see a key error from `test_observation`, just looks like it couldn't find the expected FITS file? Probably error in the glob patterns, or something like that.

AnthonyHorton:
> The hang I suspect is the result of a file name collision, but we'd have to run the test suite locally using your branch in order to see the logs from the remote camera to work out what is happening there.

Really should implement a proper timeout for the autofocus. If an exception gets raises in the thread the caller will end up waiting forever.

huntsman/tests/test_camera.py
danjampro: I was going with the assumption that `image_dir` points to the image directory on the control computer and `image_dir_local` refers to the local device storage. So a `pyro.Camera` instance should have `self.config['directories']['images'] = images_dir`.

Although, for these tests, it doesn't matter either way because both directories are "local".

At the moment `pyro.Camera` instances don't get their config from the config server, but use  `huntsman.utils.load_config` with the default of `huntsman.yaml`. This should probably be changed...

danjampro:
> .> @AnthonyHorton Hmmm.... They were working alright before I merged with #158.
> 
> > I had to undo one of your changes to get `test_autofocus_coarse` working. Specifically, I removed the hardcoded images_dir directory in test_camera.patterns fixture. But now `test_autofocus_fine` hangs and for some reason I'm getting a key error in `test_observation`, which isn't making much sense to me at the moment...
> 
> Yes, that hard coded images directory was the other half of my method of making the remote cameras use `/tmp`, post sshfs. Your new way is better.
> 
> The hang I suspect is the result of a file name collision, but we'd have to run the test suite locally using your branch in order to see the logs from the remote camera to work out what is happening there. I don't see a key error from `test_observation`, just looks like it couldn't find the expected FITS file? Probably error in the glob patterns, or something like that.

Perhaps you are right about the filename collision... If I skip the coarse test the fine one works, but then the next one fails. Its odd though, as you put in the call to `shutil.rmtree`.

huntsman/tests/test_camera.py
```diff
@@ -311,7 +313,7 @@ def test_autofocus_coarse(camera, patterns):
         pytest.skip("Camera does not have a focuser")
     try:
         autofocus_event = camera.autofocus(coarse=True)
-        autofocus_event.wait()
+        autofocus_event.wait(timeout=30)
```
AnthonyHorton: This timeout is way too short. This is the cause of the cryptic 'not enough data received' errors. The simulated cameras are perhaps a bit too realistic in that a simulated autofocus sequence actually takes quite a long time. Changing this to 150 seconds makes this test run just fine.

huntsman/tests/test_camera.py
```diff
@@ -311,7 +313,7 @@ def test_autofocus_coarse(camera, patterns):
...
         assert len(glob.glob(patterns['final'])) == 1
     finally:
         shutil.rmtree(patterns['base'])
```
AnthonyHorton: Something weird going on where this cleanup code seems to be deleting all the directories _after_ the next test has started to take an image, causing that test to hang. Oddly this doesn't happen in `develop`, only in this branch.

I tried just commenting out the `rmtree` line but then the following test failed because the glob pattern finds too many 'final' files (the one from that test and the one from this test).

PR #162, opened 2020-02-12
panoptes-utils from current github

wtgee:
Until we can do a new release of `panoptes-utils` we need to module from develop.


danjampro (approved):
LGTM.

PR #163, opened 2020-02-12
Add Pyro serialisers for POCS exceptions

AnthonyHorton:
The aim of this PR is to enable Pyro to (de)serialise POCS exceptions (e.g. `pocs.utils.error.PanError`, `Timeout`, `NotFound`, `NotSupported`, etc.) so that if a remote camera raises one of these exceptions it will get re-raised on the main control computer. Currently what happens in this situation is that a `Pyro4.error.SerializeError` gets raised instead, and we only see the type of the remote exception and neither the message nor traceback. We have to then go digging in the remote camera logs to find out what actually happened.

This is a follow on from #158, which added (de)serialisers for astropy Quantities.

I have also taken the opportunity to move all the custom serialisers code to `huntsman.utils.pyro`, added a test server, and added tests of the custom serialisers using that test server.

### DIFF
--- a/huntsman/tests/test_aardvark.py
+++ b/huntsman/tests/test_aardvark.py
@@ -5,6 +5,10 @@ import os
 
 import pytest
 import Pyro4
+import Pyro4.errors
+from astropy import units as u
+
+from pocs.utils import error
 
 from huntsman.utils import get_own_ip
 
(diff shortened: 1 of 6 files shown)


huntsman/tests/test_aardvark.py
```diff
@@ -33,7 +37,49 @@ def test_name_server(name_server):
...
+
+def test_quantity_argument(test_proxy):
+    test_proxy.quantity_argument(550 * u.nm)
+    test_proxy.quantity_argument(1.21 * u.GW)
+
+
+def test_quantity_return(test_proxy):
+    q = test_proxy.quantity_return()
+    assert isinstance(q, u.Quantity)
+    assert q == 42 * u.km * u.ng / u.Ms
```
wtgee, line 54: I don't understand why `quantity_return` with no params would be equal to `42 * u.km * u.ng / u.Ms`

huntsman/tests/test_aardvark.py
```diff
@@ -33,7 +37,49 @@ def test_name_server(name_server):
...
+def test_pocs_subclass(test_proxy):
+    with pytest.raises(error.Timeout):
+        test_proxy.raise_timeout()
+    with pytest.raises(error.NotSupported):
+        test_proxy.raise_notsupported()
+    with pytest.raises(error.IllegalValue):
+        test_proxy.raise_illegalvalue()
+
+
+def test_undeserialisable(test_proxy):
```
wtgee, line 78: undeserialisable? Is that a word? :grin:

huntsman/utils/pyro/__init__.py
```diff
@@ -1,10 +1,101 @@
...
 
 from huntsman.camera.pyro import CameraServer
 from huntsman.utils import get_own_ip, sshfs, DummyLogger
 from huntsman.utils.config import load_device_config
 
+# Enable local display of remote tracebacks
+sys.excepthook = Pyro4.util.excepthook
+
+# Serialisers/deserialisers
+name_pattern = re.compile(r"error\.(\w+)'>$")
```
wtgee: `name_pattern` seems like an odd name for this variable. `error_pattern`?

huntsman/tests/test_aardvark.py
wtgee: I see below what this is doing. I guess since this is all internal to testing it's not terribly important.

wtgee (approved):
LGTM

huntsman/tests/test_aardvark.py
AnthonyHorton: It is now!

huntsman/utils/pyro/__init__.py
AnthonyHorton: I guess I was thinking that this is the pattern that captures the name of the error from the string representation of its class, so `name_pattern`.  `error_pattern` is more descriptive, though.

AnthonyHorton: I feel like there must be a more elegant way to do what I'm using this regular expression for. I know Python has some powerful introspection features, so parsing `str(obj.__class__)` myself seems like something I shouldn't have to do. Do any alternatives spring to mind?

PR #164, opened 2020-02-13
Exclude tests & hardware dependent code from test coverage

AnthonyHorton:
Add a `.codecov.yml` file to exclude the tests themselves and code that cannot be run without real hardware from the test coverage calculations, as POCS does.


wtgee (approved):
LGTM! :grin: :rocket: :fireworks:

PR #165, opened 2020-02-13
Re-enable test_run

AnthonyHorton:
Removes the skip mark on `test_run` from `huntsman/tests/test_huntsman.py` and fixes a bug in the `calibrating.py` state that was causing it to fail.

### DIFF
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -106,7 +106,7 @@ def on_enter(event_data):
         pocs.say("Coarse focusing all cameras before starting observing for the night")
         autofocus_events = pocs.observatory.autofocus_cameras(coarse=True)
         pocs.logger.debug("Started focus, going to wait")
-        pocs.wait_for_events(autofocus_events.values(), 600)  # Longer timeout?
+        pocs.wait_for_events(list(autofocus_events.values()), 600)  # Longer timeout?
 
     except Exception as e:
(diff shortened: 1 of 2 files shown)


states/huntsman/calibrating.py
```diff
@@ -106,7 +106,7 @@ def on_enter(event_data):
         pocs.say("Coarse focusing all cameras before starting observing for the night")
         autofocus_events = pocs.observatory.autofocus_cameras(coarse=True)
         pocs.logger.debug("Started focus, going to wait")
-        pocs.wait_for_events(autofocus_events.values(), 600)  # Longer timeout?
+        pocs.wait_for_events(list(autofocus_events.values()), 600)  # Longer timeout?
```
wtgee, line 109: So we shouldn't need the explicit `list` any more, is it not working if you leave it out?

wtgee (approved):

states/huntsman/calibrating.py
AnthonyHorton: We still need it for now. You fixed this in the `listify()` in `panoptes-utils` but `pocs.core.POCS.wait_for_events()` still uses the version from `pocs.utils`, which can't handle `dict_values` objects. Once POCS switches to `panoptes-utils` we can just pass the `dict` here.

wtgee: Oh, right. Maybe the fourth time you mention that I'll actually remember. :laughing:

PR #168, opened 2020-02-13
Use panoptes-utils in favor of pocs.utils.

wtgee:
First test is to see if we can just use the `docker` branch of POCS to see if things blow up or Just Work™


danjampro:
Is this still an active PR @wtgee?

wtgee:
Yes, still some work on other side to be completed first.

Will give update at this week's meeting about status

wtgee:
Replaced by #196

PR #169, opened 2020-02-14
Restructure huntsman.utils.pyro to fix circular import issues

AnthonyHorton:
PR #163 introduced some issues with circular imports (e.g. `from huntsman.camera.pyro import CameraServer` in a new session would fail). This PR restructures the `huntsman.utils.pyro` module into separate submodules to avoid these issue.


huntsman/utils/pyro/camera_server.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Runs a Pyro camera server.
+
+    The camera server should be run on the camera control computers of distributed cameras. The
+    camera servers should be started after the name server, but before POCS.
+
+    Args:
+        ignore_local (bool, optional): If True use the default
+        $HUNTSMAN_POCS/conf_files/pyro_camera.yaml only. If False will allow
+        $HUNTSMAN_POCS/conf_files/pyro_camera_local.yaml to override the default configuration.
```
wtgee:
```suggestion
            $HUNTSMAN_POCS/conf_files/pyro_camera_local.yaml to override the default configuration.
```

huntsman/utils/pyro/camera_server.py
```diff
@@ -0,0 +1,70 @@
...
+        Default False.
```
wtgee:
```suggestion
            Default False.
```

huntsman/utils/pyro/camera_server.py
```diff
@@ -0,0 +1,70 @@
...
+    Pyro4.config.SERVERTYPE = "multiplex"
+
+    # Load the config file
+    config = load_device_config(logger=logger, **kwargs)
+
+    # Specify address
+    host = config.get('host', None)
+    if not host:
+        host = get_own_ip(verbose=True)
+    port = config.get('port', 0)
```
wtgee, line 46: Is port `0` a good default?

huntsman/utils/pyro/camera_server.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with Pyro4.Daemon(host=host, port=port) as daemon:
+        try:
+            name_server = Pyro4.locateNS()
+        except errors.NamingError as err:
+            logger.error('Failed to locate Pyro name server: {}'.format(err))
+            exit(1)
```
wtgee: Import `sys`.

```suggestion
            sys.exit(1)
```

See https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used

danjampro (approved):
LGTM!

wtgee:
I am still reviewing this, please don't merge yet.

huntsman/utils/pyro/camera_server.py
```diff
@@ -0,0 +1,70 @@
...
+
+        logger.info('Found Pyro name server.')
+        uri = daemon.register(CameraServer)
+        logger.info('Camera server registered with daemon as {}'.format(uri))
+        name_server.register(config['name'], uri, metadata={"POCS",
+                                                            "Camera",
+                                                            config['camera']['model']})
```
wtgee, line 60: What does this metadata refer to or do?

wtgee (approved):
My only real comment would be that we have a mix of spellings going on for serializer, even within a signle file.  Pyro and POCS are both using serializers so I would argue for that. Not strictly required but might would probably be good to be consistent.

huntsman/utils/pyro/camera_server.py
AnthonyHorton: It is for the Pyro Daemon. `port=0` is the default, and means pick a random (high valued) one.

https://pyro4.readthedocs.io/en/stable/servercode.html#Daemon

I could set it to `None` instead to make it more obvious that it's not literally using port 0.

wtgee: Having it as `None` is a little more intuitive (to me at least).

huntsman/utils/pyro/camera_server.py
AnthonyHorton: Objects that register with a `Pyro` name server can, optionally, register some metadata about themselves along with their name and URI. The metadata is just a `set` of strings. The main (only?) use of this metadata is to allow clients to query the name server for objects that contain one or more specified items in their metadata. `huntsman.camera.list_distributed_cameras()` uses this to get the details of registered objects that have both "POCS" and "Camera" in their metadata. That prevents it from getting the config server too, for example.

huntsman/utils/pyro/camera_server.py
AnthonyHorton: TIL.

PR #170, opened 2020-02-16
Support for filterwheels on Pyro cameras, general refactoring

AnthonyHorton:
Adds a filterwheel interface to the distributed camera system, following largely the same approach as the focusers.

I'm doing something new with remote access to Events. Would appreciate some feedback on this.

Edit: Scope expanded, just a little. Now:

- Adds filterwheel interface to the distributed camera system
- Introduces a `RemoteEvent` class, providing an interface to `threading.Event`s used by the remote cameras and their subcomponents.
- Switches all asynchronous operations (`take_exposure`, `autofocus` & filterwheel moves) to use the remote events instead of using Pyro's asynchronous results mechanism.
- No longer spawning local threads (`async_wait`) for remote asynchronous operations, allowing remote exceptions to be re-raised locally in the main thread.
- Removes now redundant parameter/status checks from local `take_exposure` and `autofocus` methods because the same checks on the remote camera are now able to raise exceptions locally.
- Added backup local timeouts for remote `take_exposure` and `autofocus` to make it less likely calling code ends up waiting forever for a failed operation. 
- Replaces large number of remote attribute getter/setter methods on the `CameraServer` with generic `get()` and `set()` methods.
- Removes unused and non-functional `key` parameter from camera server startup script and function.
- Adds fall back to 'localhost' to `load_device_config` to simplify running a camera server locally for testing, no longer necessary to override the default config files.


huntsman/camera/pyro.py
```diff
@@ -207,23 +211,23 @@ def take_exposure(self,
...
-                                      'exposure',
-                                      self._exposure_event,
-                                      self._timeout))
+        # Start a thread that will set an event once exposure has completed
+        exposure_thread = Thread(target=self._async_wait,
+                                 args=(exposure_result,
+                                       'exposure',
+                                       self._exposure_event,
+                                       get_quantity_value(seconds, u.s) +
+                                       self.readout_time + self._timeout))
```
wtgee: Maybe break this out into a variable before hand just so it's explicit: `total_exptime = ... + ...`

Also make me wonder if we shouldn't just have a property or method that returns that value (different PR).

huntsman/camera/pyro.py
```diff
@@ -684,3 +688,65 @@ def focuser_move_to(self, position):
 
     def focuser_move_by(self, increment):
         return self._camera.focuser.move_by(increment)
+
+# Filterwheel methods - these are used by the remote filterwheel client,
+# huntsman.filterwheel.pyro.FIlterWheel
```
wtgee:
```suggestion
# huntsman.filterwheel.pyro.FilterWheel
```

huntsman/filterwheel/pyro.py
```diff
@@ -0,0 +1,98 @@
...
+    def is_moving(self):
+        """ Is the filterwheel currently moving """
+        return self._proxy.filterwheel_is_moving
+
+    @property
+    def is_ready(self):
+        # A filterwheel is 'ready' if it is connected and isn't currently moving.
+        return self._proxy.filterwheel_is_ready
+
+    @AbstractFilterWheel.position.getter
```
wtgee, line 41: Should this be here still? Same for next property.

huntsman/filterwheel/pyro.py
```diff
@@ -0,0 +1,98 @@
...
+        self._serial_number = self._proxy.filterwheel_uid
+
+        self.logger.debug(f"{self} connected.")
+
+##################################################################################################
+# Private methods
+##################################################################################################
+
+    def _move_to(self, position):
+        self._proxy.filterwheel_move_to(position)
```
wtgee, line 77:
```suggestion
        self._proxy.filterwheel_move_to(position)

```

Files should end with a blank line.

huntsman/camera/pyro.py
```diff
@@ -684,3 +688,65 @@ def focuser_move_to(self, position):
...
+        self._camera.filterwheel._move_event.set()
+
+    def filterwheel_event_clear(self):
+        self._camera.filterwheel._move_event.clear()
+
+    def filterwheel_event_is_set(self):
+        return self._camera.filterwheel._move_event.is_set()
+
+    def filterwheel_event_wait(self, timeout=None):
+        return self._camera.filterwheel._move_event.wait(timeout)
```
wtgee:
```suggestion
        return self._camera.filterwheel._move_event.wait(timeout)

```

End with a blank line.

wtgee (commented):
LGTM.  I like the class based event access although I'm curious what advantages you are hoping it gives. Seems slightly more organized to me but I'm not sure if that was the motivation.

AnthonyHorton:
> LGTM. I like the class based event access although I'm curious what advantages you are hoping it gives. Seems slightly more organized to me but I'm not sure if that was the motivation.

A few thoughts motivating this. The alternative approach that I used for the remote camera `take_exposure` and `autofocus` methods was to force the calls to the remote camera to be blocking, and use the Pyro async results stuff and a `_async_wait()` method to poll from the client side and set an `Event` there. When faced with the prospect of doing the same for the filterwheel moves it occurred to me that there was a bunch of probably unnecessary duplication going on there. The camera, focuser & filterwheel class already have their own polling threads for setting Events when operations complete, and all I needed in order to make use of those was to have a way to access those remote Events. Worse, by making the remote calls blocking I was preventing the CameraServer from responding to any other calls (e.g. status checks) while an operation was in progress. This would have effects such as making it harder to check to check whether a filterwheel was already in progress when requesting another one (a call to `is_moving()` on the server would not return until any current move had completed).

huntsman/filterwheel/pyro.py
AnthonyHorton: They are needed. I can (and do) use the setters from the base class but need to override the getters.

AnthonyHorton:
Scope on this one expanded rather a lot as I attempted to fix some troublesome bugs. Now remote Events are used for all asynchronous operations (exposures, autofocus, filterwheel moves) and the Pyro asynchronous results machinery has been entirely removed. This has improved functionality, with a larger percentage of possible errors generating the same exceptions that they would do for a local camera. I've also made access to remote camera and subcomponent attributes and events generic, which has significantly reduced the amount of code in the CameraServer class and made it easier and cleaner to poke around inside the remote camera to artificially induce errors for unit tests.

scripts/pyro_camera_server.py
```diff
@@ -22,4 +22,5 @@
     args = parser.parse_args()
 
     # Run the camera server
-    run_camera_server(args.ignore_local, key=args.key)
+    run_camera_server(ignore_local=args.ignore_local,
+                      key=args.key)
```
wtgee: This is fine but it kind of looks like your auto-formatter is set at 80 chars.

huntsman/utils/pyro/serializers.py
```diff
@@ -56,8 +56,23 @@ def dict_to_astropy(class_name, d):
...
+
 SerializerBase.register_class_to_dict(u.Quantity, astropy_to_dict)
 SerializerBase.register_dict_to_class("astropy_yaml", dict_to_astropy)
 
 SerializerBase.register_class_to_dict(error.PanError, panerror_to_dict)
 SerializerBase.register_dict_to_class("PanError", dict_to_panerror)
+
+# These two are only here as a temporary workaround for some typos in POCS
+SerializerBase.register_class_to_dict(ValueError, value_error_to_dict)
+SerializerBase.register_dict_to_class("ValueError", dict_to_value_error)
```
wtgee, line 78: I don't understand the comment and associated code. What are the typos in POCS?

huntsman/utils/pyro/event.py
```diff
@@ -0,0 +1,27 @@
+from threading import Event
+
+event_types = {"camera",
+               "focuser",
+               "filterwheel"}
+
+
+class RemoteEvent(Event):
+    """Interface for threading.Events of a remote camera or its subcomponents (e.g. filterwheel)
+    """
```
wtgee: Maybe list all the types in the docstring itself. Either in description or providing params description.
```suggestion
    """Interface for threading.Events of a remote camera or its subcomponents.
    
    Current supported types are: `camera`, `focuser`, `filterwheel`.
    """
```

huntsman/utils/pyro/camera_server.py
```diff
@@ -9,7 +9,10 @@
 from huntsman.camera.pyro import CameraServer
 
 
-def run_camera_server(ignore_local=False, unmount_sshfs=True, logger=None,
+def run_camera_server(ignore_local=False,
+                      key=None,
```
wtgee: I assumed this was an ssh key but it looks like it is a config key item?  Maybe be more explicit and add to docstring.

wtgee: Or maybe just pass it through via `kwargs` if `run_camera_server` itself doesn't need it. Although that's a bit hiding it.

huntsman/tests/test_camera.py
```diff
@@ -235,7 +236,7 @@ def test_exposure_scaling(camera, tmpdir):
     else:
         fits_path = str(tmpdir.join('test_exposure_scaling.fits'))
         camera.take_exposure(filename=fits_path, dark=True, blocking=True)
-        image_data, image_header = fits_utils.getdata(fits_path, header=True)
+        image_data, image_header = fits.getdata(fits_path, header=True)
```
wtgee, line 239: Oh, hm. I should support returning the header from the `fits_utils` although I'm kind of surprised this just doesn't work.

huntsman/camera/pyro.py
```diff
@@ -177,73 +185,51 @@ def take_exposure(self,
...
-                                      self._exposure_event,
-                                      self._timeout))
-        exposure_thread.start()
+        self._proxy.take_exposure(seconds=seconds,
+                                  filename=filename,
+                                  dark=bool(dark),
+                                  *args,
+                                  **kwargs)
 
+        max_wait = get_quantity_value(seconds, u.second) + self.readout_time + self._timeout
```
wtgee, line 216: Shouldn't it be `max((seconds + readout), self._timeout)`?  That is, the length of exposure and readout (and maybe some padding) or the timeout?

huntsman/camera/pyro.py
```diff
@@ -177,73 +185,51 @@ def take_exposure(self,
...
         if blocking:
-            self._exposure_event.wait()
+            success = self._exposure_event.wait(timeout=max_wait)
+            if not success:
+                self._timeout_response("exposure")
+        else:
+            # If the remote exposure fails after starting in such a way that the event doesn't
+            # doesn't get set then calling code could wait forever. Have a local timeout thread
+            # to be safe.
```
wtgee: Removing extra "doesn't"

```suggestion
            # If the remote exposure fails after starting in such a way that the event doesn't
            # get set then calling code could wait forever. Have a local timeout thread to be safe.
```

huntsman/camera/pyro.py
```diff
@@ -177,73 +185,51 @@ def take_exposure(self,
...
-                  take_dark=None,
-                  merit_function='vollath_F4',
-                  merit_function_kwargs={},
-                  mask_dilations=None,
-                  coarse=False,
-                  make_plots=False,
-                  blocking=False,
-                  timeout=None,
-                  *args, **kwargs):
+    def autofocus(self, blocking=False, *args, **kwargs):
```
wtgee, line 221: Is it still appropriate to have all the params documented here?

huntsman/camera/pyro.py
```diff
@@ -275,55 +261,44 @@ def autofocus(self,
...
+        self.logger.debug(f'Starting autofocus on {self}.')
+
+        # Remote method call to start the exposure
+        self._proxy.autofocus(*args, **kwargs)
 
+        # Proxy for remote _autofocus_event
+        self._autofocus_event = RemoteEvent(self._proxy, event_type="focuser")
+
+        # In general it's very complicated to work out how long an autofocus should take
+        # because parameters can be set here or come from remote config. For not just make
```
wtgee:
```suggestion
        # because parameters can be set here or come from remote config. For now just make
```

huntsman/camera/pyro.py
```diff
@@ -275,55 +261,44 @@ def autofocus(self,
...
+            if not success:
+                self._timeout_response("autofocus")
+        else:
+            # If the remote autofocus fails after starting in such a way that the event doesn't
+            # doesn't get set then calling code could wait forever. Have a local timeout thread
+            # to be safe.
+            timeout_thread = Timer(interval=max_wait,
+                                   function=self._timeout_response,
+                                   args=("autofocus",))
+            timeout_thread.start()
```
wtgee: This is almost a repeat of (new) lines 210-218, might be worth wrapping up into a helper method. Might no. :laughing:

huntsman/camera/pyro.py
```diff
@@ -335,30 +310,22 @@ def _readout(self, filename=None):
...
-        else:
-            if event is not None:
-                event.set()
-            msg = "Timeout while waiting for {} on {}".format(name, self.port)
-            raise error.Timeout(msg)
+    def _timeout_response(self, timeout_type):
+        # This could do more thorough checks for success, e.g. check is_exposing property,
+        # check for existence of output file, etc. It's supposed to be a last resort though,
+        # and most problems should be caught elsewhere.
+        relevant_event = getattr(self, f"_{timeout_type}_event")
```
wtgee: Could have an `AttributeError` here without a default. I'll leave it to you as to whether to set default or let error happen. I see the note above so not too worried about it.

huntsman/camera/pyro.py
```diff
@@ -335,30 +310,22 @@ def _readout(self, filename=None):
...
+        is_set = True
+        try:
+            is_set = relevant_event.is_set()
+        except Pyro4.errors.CommunicationError:
+            # Can happen is everything has finished and shutdown before timeout,
+            # e.g. when running tests.
+            pass
```
wtgee:
```suggestion
        # Exception can happen if everything has finished and shutdown before 
        # timeout, e.g. when running tests.
        with suppress(Pyro4.errors.CommunicationError):
            is_set = relevant_event.is_set()
```

huntsman/camera/pyro.py
```diff
@@ -436,87 +406,20 @@ def __init__(self, config_files=None):
...
+# them all with generic get and set methods.
 
-    @property
-    def readout_time(self):
-        return self._camera.readout_time
+    def get(self, property_name, subcomponent=None):
+        obj = self._camera
+        if subcomponent:
+            obj = getattr(obj, subcomponent)
+        return getattr(obj, property_name)
```
wtgee, line 405: I like the cleanup on this side of things but am a little worried about the string based calling from the user's perspective.  See note in review comments.

huntsman/camera/pyro.py
wtgee: Can this just be removed now?

wtgee (approved):
This looks good, my only concern would be that with the move to accessing a lot of the properties on  the proxy via `get`/`set`, you are now using strings instead of attributes (or method names), which I think could be more error prone. I think it could also be cleaned up via some clever currying or use of [`functools.partialmethod`](https://docs.python.org/3.8/library/functools.html#functools.partialmethod), but that would require more thought so should be left for separate PR.

Can you also update PR description with included changes?

danjampro (approved):

huntsman/camera/pyro.py
AnthonyHorton: Hmmmm, we probably need to clarify the role of the `Camera` class `timeout` attribute. In the POCS `Camera` base class `take_exposure` starts a `threading.Timer` with duration of `seconds`, which then starts the polling loop which runs for a maximum period of `timeout` seconds. So in that case the maximum elapsed time to wait for an exposure is `seconds + timeout`, the `readout_time` property is totally ignored (!), and `timeout` should be set to the camera's nominal readout time plus a little padding.

Here I'm using the `timeout` attribute of the Pyro `Camera` (note, it will in general be different from the corresponding remote camera's `timeout`) as the padding and so set the maximum elapsed time to wait for the exposure to `seconds + readout_time + timeout`. This makes more sense to me than what POCS is currently doing, but we should be consistent here.

Maybe it can be cleared up with renaming the current `._timeout` attribute to something like `._timeout_padding` and then the above line becomes `timeout = get_quantity_value(seconds, u.second) + self.readout_time + self.timeout_padding`.

huntsman/camera/pyro.py
AnthonyHorton: I thought so, as while this is just a duplication of the docstring from the `Camera` base class (and the `Focuser` base class) it is a public method of a potentially user facing class so having the docstring here might help with code editor inline help/tooltips/autocomplete/etc.

huntsman/camera/pyro.py
AnthonyHorton: I'm happy with an `AttributeError` happening here. If the relevant Event is missing something has gone badly wrong and and raising an exception is the appropriate response.

huntsman/camera/pyro.py
AnthonyHorton: Dangling more opportunities for reducing line count in front of me... OK, done.

huntsman/camera/pyro.py
AnthonyHorton: Done, but with some other overlapping changes so not using this suggested commit.

huntsman/camera/pyro.py
AnthonyHorton: Not quite. Don't need separate functions to get/set all the properties any more, but the method calls are different.

huntsman/tests/test_camera.py
AnthonyHorton: Yeah, this was showing up in test failures. This test is usually skipped, but something broke the skipping while I was working on this PR and I found that `fits_utils` doesn't have `get_data` (at least not the one from POCS `develop`).

huntsman/utils/pyro/camera_server.py
AnthonyHorton: Yes, config key. This parameter got added along with the config server stuff, but when I tried to actually use it to run a local camera server for testing purposes I discovered it didn't actually work. My changes in this file were the result of my attempt to fix it so that I could use a local camera server to work out some tricky bugs in the Pyro camera stuff. After a fair bit of wasted time poking at this stuff I discovered that it wasn't going to work anyway because there is no way to pass the `key` parameter to the `CameraServer` instance, and that will always try to use a config for the host's IP address regardless. That's why the config server test fixture in `conftest` inserts a device config with the host's IP address on the fly, and why I ended up added the `localhost` fallback in `huntsman.utils.config.load_device_config`. I should just remove all traces of the `key` parameter since it's not used and can't work.

huntsman/utils/pyro/serializers.py
AnthonyHorton: In POCS there are a few instance of 

`raise ValueError("Blah, blah, blah, {}!", obj)`

where it should be

`raise ValueError("Blah, blah, blah, {}!".format(obj))`

or

`raise ValueError(f"Blah, blah, blah, {obj}!")`, e.g. 

https://github.com/panoptes/POCS/blob/22532dab81745aa5b805d809b4c494bd5495bc39/pocs/focuser/focuser.py#L242 and https://github.com/panoptes/POCS/blob/22532dab81745aa5b805d809b4c494bd5495bc39/pocs/focuser/focuser.py#L249 

These don't cause problems with local cameras (beyond funny looking exceptions) but it was causing test failures with the remote cameras because Pyro was trying to serialise `ValueErrors` that contained, in their `args` list, a reference to an object that it couldn't serialise. It took me a while to track down the root cause of the opaque 'circular reference' exceptions that Pyro was throwing as a result! My workaround in order to get this PR working was to add a custom serialiser/deserialiser for `ValueError` that simply casts all of the `args` to `str`, thereby making sure that `Pyro` can handle them. If we fix this in POCS then I can take this out again.

scripts/pyro_camera_server.py
AnthonyHorton: Reverted, and more, because I've taken the key parameter out.

I do have an issue with my linter plugin though. Whatever I do I can't get it to accept a custom line length, which means it flags issues all over POCS and huntsman-pocs code, making it hard to spot real issues.

AnthonyHorton:
> This looks good, my only concern would be that with the move to accessing a lot of the properties on the proxy via `get`/`set`, you are now using strings instead of attributes (or method names), which I think could be more error prone. I think it could also be cleaned up via some clever currying or use of [`functools.partialmethod`](https://docs.python.org/3.8/library/functools.html#functools.partialmethod), but that would require more thought so should be left for separate PR.

Yep, the generic, attribute name based `get()` and `set()` are clunky, but I make myself feel better about them by reminding myself that the `CameraServer` class is not intended to be a user facing interface. It exists only as a wrapper around the remote `Camera` instance to enable the Pyro network layer. You're not supposed to interact with the `CameraServer` directly, you're supposed to use the Pyro `Camera`, `Focuser` and `Filterwheel` classes instead. Those three classes should be the only code calling `CameraServer` methods.

> Can you also update PR description with included changes?

Done. Quite a long list now!

huntsman/utils/pyro/serializers.py
AnthonyHorton: https://github.com/panoptes/POCS/issues/966

PR #172, opened 2020-02-17
pep8 cleanup

wtgee:
95% of this is just applying pep8 to the files.  Some additional removal of old code.  A few minor fixes.

### DIFF
--- a/huntsman/camera/__init__.py
+++ b/huntsman/camera/__init__.py
@@ -152,10 +152,8 @@ def create_distributed_cameras(camera_info, logger=None):
     for cam_name, cam_uri in camera_uris.items():
         logger.debug('Creating camera: {}'.format(cam_name))
         cam = PyroCamera(port=cam_name, uri=cam_uri)
-        is_primary = ''
         if primary_id == cam.uid or primary_id == cam.name:
             cam.is_primary = True
-            is_primary = ' [Primary]'
 
         logger.debug(f"Camera created: {cam}")
 
--- a/huntsman/tests/test_camera.py
+++ b/huntsman/tests/test_camera.py
@@ -29,6 +29,8 @@ params = [PyroCamera]
 ids = ['pyro']
 
 # Ugly hack to access id inside fixture
+
+
 @pytest.fixture(scope='module', params=zip(params, ids), ids=ids)
 def camera(request, images_dir_control, camera_server):
     if request.param[0] == PyroCamera:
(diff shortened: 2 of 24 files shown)


huntsman/camera/__init__.py
```diff
@@ -152,10 +152,8 @@ def create_distributed_cameras(camera_info, logger=None):
     for cam_name, cam_uri in camera_uris.items():
         logger.debug('Creating camera: {}'.format(cam_name))
         cam = PyroCamera(port=cam_name, uri=cam_uri)
-        is_primary = ''
         if primary_id == cam.uid or primary_id == cam.name:
             cam.is_primary = True
-            is_primary = ' [Primary]'
```
wtgee, line 158: @AnthonyHorton did you intend to alter the camera string property? This was unused.

huntsman/tests/test_camera.py
```diff
@@ -233,7 +235,7 @@ def test_exposure_scaling(camera, tmpdir):
     else:
         fits_path = str(tmpdir.join('test_exposure_scaling.fits'))
         camera.take_exposure(filename=fits_path, dark=True, blocking=True)
-        image_data, image_header = fits.getdata(fits_path, header=True)
+        image_data, image_header = fits_utils.getdata(fits_path, header=True)
```
wtgee, line 238: This was an error. Not sure how test was passing.

states/huntsman/dithering.py
```diff
@@ -25,4 +25,4 @@ def on_enter(event_data):
             pocs.next_state = 'observing'
 
     except Exception as e:
-        pocs.logger.warning("Problem with preparing: {}".format())
+        pocs.logger.warning("Problem with preparing: {}".format(e))
```
wtgee, line 28: This exception would have been raising an exeption.

huntsman/tests/test_camera.py
AnthonyHorton: This tests usually gets skipped as it depends on a camera having a bit depth other than 16 bits/pixel and the huntsman-pocs test suite only tests with a single type of simulated camera (Pyro, SDK, 16 bit, with focuser). Running the tests on a variety of simulated cameras like in POCS is on my todo list.

huntsman/camera/__init__.py
AnthonyHorton: ISTR there being an `is_primary` variable being actually  used in this area of code at some point. I think these lines are orphaned code fragments from times past... Deleting them is the merciful thing to do.

wtgee: This exists in POCS when the camera is created:

https://github.com/panoptes/POCS/blob/5dbd821b582eedac992da9124defcacc03b0e09a/pocs/camera/__init__.py#L183-L190

AnthonyHorton (approved):
LGTM. Might want to throw in this `is` -> `=` while you're doing it, though: https://github.com/AstroHuntsman/huntsman-pocs/blob/591e16b12766dfa6c178be5dfdcbaf0639b26b44/states/huntsman/calibrating.py#L26

PR #178, opened 2020-03-03
Fix NGAS push

danjampro:
Fixes a bug where if `filename_ngas` in `pyro.Camera._NGASpush` is not specified, `filename` would be modified rather than `filename_ngas`.


lspitler (approved):
LGTM

PR #179, opened 2020-04-17
Fix calibrating state

danjampro:
This PR fixes a bug in the calibrating state where if the safety check failed after the flats were taken (therefore calling the parking state), the calibrating state would ignore this and continue to do the coarse focusing. After the focusing finishes, calibrating tries to go to the scheduling state, but is unable to do so as the machine is actually in parking/parked rather than calibrating at this stage. In future we will move the coarse autofocus code to its own state.

I also heavily modified (and shrunk) the code to make proper use of the POCS horizons, which are specified in the config file, rather than hardcoded horizons.

### DIFF
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -6,7 +6,7 @@ from pocs.utils import current_time
 
 def _wait_for_twilight(pocs, horizon):
     '''
-    Wait for twighlight if safe to do so.
+    Wait for twilight if safe to do so.
     '''
     delay = pocs._safe_delay
 
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -71,7 +71,7 @@ def on_enter(event_data):
     # Wait for twilight
     coarse_focus_timeout = 600  # Put this in a config file
     if not _wait_for_twilight(pocs, horizon='focus'):
-        print('Exiting calibrating state because it is no longer safe.')
+        pocs.logger.info('Exiting calibrating state because it is no longer safe.')
         return
 
     try:
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -46,10 +46,6 @@ def on_enter(event_data):
     pocs = event_data.model
     pocs.next_state = 'parking'
 
-    # Leave the state if not safe to take calibrations
-    if not pocs.is_safe(horizon='flat'):
-        print('Exiting calibrating state because it is no longer safe.')
-        return
 
     if pocs.observatory.take_flat_fields:
 
--- a/states/huntsman/calibrating.py
+++ b/states/huntsman/calibrating.py
@@ -4,116 +4,87 @@ from astropy import units as u
 from pocs.utils import current_time
 
 
-def wait_for_sun_alt(pocs,
-                     min_altitude=None,
-                     max_altitude=None,
-                     delay=None,
-                     message="Waiting for Sun altitude. Current: {}"):
-    """
-    Wait for the altitude of the Sun to be within given limits
+def _wait_for_twilight(pocs, horizon):
+    '''
+    Wait for twighlight if safe to do so.
+    '''
+    delay = pocs._safe_delay
 
-    Args:
-        pocs :
-        min_altitude (astropy.units.Quantity, optional):
-        max_altitude (astropy.units.Quantity, optional):
-        message (str):
-    """
-    if min_altitude is None and max_altitude is None:
-        raise ValueError("At least one of min_altitude & max_altitude must be given")
-    if min_altitude is None:
-        min_altitude = -90
-    if max_altitude is None:
-        max_altitude = 90
-    if isinstance(min_altitude, u.Quantity):
-        min_altitude = min_altitude.to(u.degree).value
-    if isinstance(max_altitude, u.Quantity):
-        max_altitude = max_altitude.to(u.degree).value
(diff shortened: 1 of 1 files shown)


states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
-            # Check simulator for 'night'
-            if 'night' in pocs.config['simulator']:
-                pocs.logger.info(f'Using night simulator, pretending sun down.')
-                break
-            else:
-                pocs.say(message.format(sun_pos.value))
-                pocs.sleep(delay=delay)
+def _wait_for_twilight(pocs, horizon):
+    '''
+    Wait for twighlight if safe to do so.
```
wtgee:
```suggestion
    Wait for twilight if safe to do so.
```

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
-        pocs.logger.warning("Problem with flat fielding: {}".format(e))
+        try:
+            # Identify filter types
+            narrow_band_cameras, broad_band_cameras = _get_cameras(pocs)
+
+            # Take calibration frames
+            pocs.say("Starting narrow band flat fields")
+            pocs.observatory.take_evening_flats(camera_list=narrow_band_cameras)
+
+            pocs.say("Staring broad band flat fields")
```
wtgee:
```suggestion
            pocs.say("Starting broad band flat fields")
```

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
+            pocs.observatory.take_evening_flats(camera_list=broad_band_cameras)
+
+        except Exception as err:
+            pocs.logger.warning(f'Problem with flat fielding: {err}')
+
+    # This code needs to be moved to its own state
+    # Wait for twilight
+    coarse_focus_timeout = 600  # Put this in a config file
+    if not _wait_for_twilight(pocs, horizon='focus'):
+        print('Exiting calibrating state because it is no longer safe.')
```
wtgee:
```suggestion
        pocs.logger.info('Exiting calibrating state because it is no longer safe.')
```

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
+    called once at the beginning and end of the night.
+    '''
+    pocs = event_data.model
     pocs.next_state = 'parking'
 
-    try:
+    # Leave the state if not safe to take calibrations
+    if not pocs.is_safe(horizon='flat'):
+        print('Exiting calibrating state because it is no longer safe.')
+        return
```
wtgee: This check shouldn't be necessary here as safety would have bee required to enter the state and nothing else has been done at this point. Should be able to just remove all this.

```suggestion
```

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
+def _wait_for_twilight(pocs, horizon):
+    '''
+    Wait for twighlight if safe to do so.
+    '''
+    delay = pocs._safe_delay
+
+    while pocs.is_safe(horizon='flat'):
+
+        if not pocs.is_dark(horizon=horizon):
+            pocs.say('Not dark enough for coarse focusing. Waiting...')
```
wtgee, line 16:
```suggestion
            pocs.say(f'Not dark enough for {horizon}. Waiting...')
```

Or something similar. This is also used for waiting on observe.

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
+
+            pocs.say("Staring broad band flat fields")
+            pocs.observatory.take_evening_flats(camera_list=broad_band_cameras)
+
+        except Exception as err:
+            pocs.logger.warning(f'Problem with flat fielding: {err}')
+
+    # This code needs to be moved to its own state
+    # Wait for twilight
+    coarse_focus_timeout = 600  # Put this in a config file
```
wtgee, line 68: Can you move this line below the safety check and grouped with the focus events?

states/huntsman/calibrating.py
```diff
@@ -4,116 +4,87 @@
...
-                                                  ).alt
-        if sun_pos.value > max_altitude or sun_pos.value < min_altitude:
-            # Check simulator for 'night'
-            if 'night' in pocs.config['simulator']:
-                pocs.logger.info(f'Using night simulator, pretending sun down.')
-                break
-            else:
-                pocs.say(message.format(sun_pos.value))
-                pocs.sleep(delay=delay)
+def _wait_for_twilight(pocs, horizon):
```
wtgee, line 7: Not terribly important but it's probably better named `wait_for_horizon` instead of `twilight`.

wtgee (approved):
Great, nice to have that clean up.

It would of course be better if we could write some tests for this as there are some tricky conditions of getting into and out of state based on sun position. I am happy to rely on the `is_dark` to encapsulate all that logic but I guess I guess a little worried with all the possible logical paths here.

states/huntsman/calibrating.py
danjampro: I will be moving all this in an imminent PR so no need to worry for now.

states/huntsman/calibrating.py
danjampro: This will be changed in the upcoming coarse autofocus PR

states/huntsman/calibrating.py
danjampro: This gets modified in an upcoming PR

PR #180, opened 2020-04-18
Coarse focusing state

danjampro:
Adds a coarse focusing state that goes in between calibrating and scheduling in the evening only. In the morning, the calibration state goes to parking. Reuses most of the old coarse focus code that was originally in the calibration state.

Tested on Huntsman hardware. Correctly goes though states in evening. Haven't checked morning calibs yet, but that can be done later.


conf_files/huntsman.yaml
```diff
@@ -59,6 +59,8 @@ messaging:
     cmd_port: 6500
     msg_port: 6510
 state_machine: /var/huntsman/huntsman-pocs/resources/state_table/huntsman.yaml
+timeout:
+  coarse_focus: 600
```
AnthonyHorton: Does this work? There's an autofocus timeout hard-coded into the Pyro `Camera` class but I didn't think there were timeouts implemented on the control computer side.

Edit: OK, found where you are using this, more comments there.

resources/state_table/huntsman.yaml
```diff
@@ -69,9 +72,13 @@ transitions:
         dest: scheduling
         trigger: schedule
     -
-        source: calibrating
+        source: coarse_focusing
```
AnthonyHorton, line 79: Shouldn't we keep the transition from `calibrating` to `scheduling` in here as well? Otherwise there is no path to get to `scheduling` after local midnight.

states/huntsman/calibrating.py
```diff
@@ -38,49 +21,48 @@ def _get_cameras(pocs):
...
-        return
-
-    pocs.next_state = 'scheduling'
+    else:
+        pocs.say('Skipping calibration frames.')
+
+    # Specify the next state
+    if _past_midnight(pocs):
+        pocs.logger.debug('Finished morning calibrations.')
+        pocs.next_state = 'parking'
```
AnthonyHorton: If you always go from `calibrating` back to `parking` after midnight then there is no way to get to `scheduling`. Should set next state to `scheduling` and leave it up to that state to decide whether to start observing right now or whether to sleep for a while.

states/huntsman/calibrating.py
```diff
@@ -38,49 +21,48 @@ def _get_cameras(pocs):
...
 
+def _past_midnight(pocs):
+    '''
+    Check if is morning, useful for going into either morning or evening flats.
+    '''
+    # Get the time of the nearest midnight to now
+    midnight = pocs.observatory.observer.midnight(current_time(), which='nearest')
+
+    # If the nearest midnight is in the past, its the morning...
+    return midnight < current_time()
```
AnthonyHorton: What we actually want is to do when deciding whether we need a coarse focus is check how long ago the last (successful) focusing was. Checking whether it's past midnight or not does not do what we want under several possible situations, e.g.:

1. If the weather is bad until past local solar midnight but then clears up then Huntsman will not do a coarse focus at all. If it happens to be the first clear night for a week then this probably won't go well!
1. If the weather is variable and goes from good to bad then back to good all before local midnight then Huntsman would do a coarse focus every time the weather comes good. This is may or may not be necessary, depending on how long the weather was bad (and how much the temperature changed in the meantime), but there's no logical reason why variable weather should result in repeated coarse focusing before local solar midnight but not afterwards.

I think the best check we could use for this (at least until we have a better idea of the relationship between temperature changes and focus drift) is the time since the last fine (not coarse) focus, and do a coarse focus if it was more than, say, a couple of hours ago. That way we will always get a coarse focus before beginning observing for the first time that night, we won't get a second one if there is a brief interruption to observing due to weather, but we will get another coarse focus if there is a lengthy break. We don't currently record the time of focusing, but it wouldn't be hard to add another member variable to the `HuntsmanObservatory` class.

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,27 @@
+"""
+State to handle the coarse focusing at the start of the night.
+"""
+
+
+def on_enter(event_data):
+    '''
+    Calibrating state. If safe to do so, take flats and darks. Should be
+    called once at the beginning and end of the night.
```
AnthonyHorton: Docstring needs updating.

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,27 @@
...
+
+    The next state is scheduling.
+    '''
+    pocs = event_data.model
+    pocs.next_state = 'parking'
+
+    # Start the autofocusing
+    coarse_focus_timeout = pocs.config['timeout']['coarse_focus']
```
AnthonyHorton: Using a fixed timeout here is probably going to cause problems once we start using narrowband filters, unless we make it *very* long. Note that there are timeouts implements at the camera level in the Pyro `Camera` class so having one at the `Observatory` level is redundant, and the camera level one is easier to make adaptive to the autofocus settings used.

AnthonyHorton (changes requested):
I think the use of a before/after midnight check is the wrong way to go here, I've discussed this more in the comments. We also need to add unit tests whenever we're changing the state machine that will cover the new transitions and associated logic.

conf_files/huntsman.yaml
danjampro: This preserves behaviour from the original state.

states/huntsman/coarse_focusing.py
danjampro: Hmm yes maybe, although I'm not sure what the best workaround is given we should probably keep using `pocs.wait_for_events` if possible, which requires a timeout to be set.

huntsman/observatory.py
```diff
@@ -386,6 +401,12 @@ def take_evening_flats(self,
                 self.logger.debug('Waiting for dark-field image')
                 time.sleep(1)
 
+    def take_morning_flats(*args, **kwargs):
+        '''
+
+        '''
+        raise NotImplementedError('Morning flats not implemented yet.')
+
```
AnthonyHorton, line 431: Does this really need to be a separate method, or could we adapt the existing `take_evening_flats()` method to work at either dusk or dawn? I haven't looked through it carefully but I don't expect that much would have to change, just a few signs would need to flip depending on which side of midnight it was.

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,29 @@
...
+        autofocus_events = pocs.observatory.autofocus_cameras(coarse=True)
+        pocs.logger.debug("Waiting for coarse focus to finish.")
+        pocs.wait_for_events(list(autofocus_events.values()), coarse_focus_timeout)
+
+        # Update last focus time
+        pocs.observatory.last_focus_time = current_time()
+
+    except Exception as err:
+        pocs.logger.warning(f"Problem with coarse autofocus: {err}.")
+
```
AnthonyHorton: The original `calibrating` state had a `_wait_for_twilight()` here before the proceeding to `scheduling`. Don't we still need that, or does the state machine handle that for us?

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+                flat_func(camera_list=camera_list)
+
+        except Exception as err:
+            pocs.logger.warning(f'Exception during twilight flat fielding: {err}')
+    else:
+        pocs.say('Skipping twilight flat fields.')
+
+    # Specify the next state
+    if pocs.observatory.require_coarse_focus():
+        pocs.next_state = 'coarse_focusing'
```
AnthonyHorton: Don't we need to wait for `is_dark(horizon='focus') == True` here? Or does the state machine handle that for us?

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
+        pocs.next_state = 'scheduling'
```
AnthonyHorton: Don't we need to wait for `is_dark(horizon='observe') == True` here? Or does the state machine handle that for us?

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,29 @@
...
+    coarse_focus_timeout = pocs.config['focusing']['coarse']['timeout']
+    try:
+        # Do the autofocusing
+        pocs.say("Coarse focusing all cameras before starting observing for the night.")
+        autofocus_events = pocs.observatory.autofocus_cameras(coarse=True)
+        pocs.logger.debug("Waiting for coarse focus to finish.")
+        pocs.wait_for_events(list(autofocus_events.values()), coarse_focus_timeout)
+
+        # Update last focus time
+        pocs.observatory.last_focus_time = current_time()
```
AnthonyHorton: I still think this is the wrong criterion. It doesn't matter if it's been 6 or however many hours since the last coarse focus provided that the focus has been continually tweaked in the meantime with fine focusing at the start of every dither sequence. Should be using the time of the last completed fine focus here, not coarse focus, and require a fine focus to have taken place in the last 2 hours, say (they normally happen every ~1 hour while observing).

Can update the `last_focus_time`at the end of the `Observatory.autofocus_cameras()` method.

AnthonyHorton (changes requested):
I do think time since last fine focus is a better criterion to use for deciding whether to coarse focus or not than time since last coarse focus.

Also these changes to the state machine need unit tests specifically to check that the new functionality is Working As Intended.

states/huntsman/coarse_focusing.py
danjampro: Handled by the state machine.

states/huntsman/twilight_flat_fielding.py
danjampro: State machine does that sort of thing.

states/huntsman/twilight_flat_fielding.py
danjampro: State machine does that sort of thing.

huntsman/observatory.py
danjampro: Not sure yet, I'm planning on tackling that in a separate PR though. This is just a placeholder that fits with the current style of the code.

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,26 @@
...
+    coarse_focus_timeout = pocs.config['focusing']['coarse']['timeout']
+    try:
+        # Do the autofocusing
+        pocs.say("Coarse focusing all cameras before starting observing for the night.")
+        autofocus_events = pocs.observatory.autofocus_cameras(coarse=True)
+        pocs.logger.debug("Waiting for coarse focus to finish.")
+        pocs.wait_for_events(list(autofocus_events.values()), coarse_focus_timeout)
+
+    except Exception as err:
+        pocs.logger.warning(f"Problem with coarse autofocus: {err}.")
```
wtgee: We want to keep observing if a problem with focusing? @AnthonyHorton

wtgee (commented):
I'll default to @AnthonyHorton suggestion on dealing with midnight but in general it is annoying. Seems like there should be cleaner, easier way.

Also, can you please start using triple double quotes for docstrings and condensing one-line docstrings to one line. :)  [One-line Docstrings](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) (the recommendation for triple double-quotes is in the paragraph before that.

states/huntsman/coarse_focusing.py
AnthonyHorton: Probably not, but that's a whole other can of worms. Properly checking for success of the autofocus and doing something sensible in case of failure has been on the todo list for a long time. Currently we don't really have a way to recover from or workaround any sort of camera/filterwheel/focuser error. If nothing else we should be disabling cameras in case of error so that Huntsman can continue observing with the remaining onrd. That's definitely a job for several other PRs though.

resources/state_table/huntsman.yaml
```diff
@@ -62,14 +65,22 @@ transitions:
...
     -
         source: ready
-        dest: calibrating
-        trigger: calibrate
+        dest: twilight_flat_fielding
+        trigger: take_twilight_flats
     -
         source: ready
         dest: scheduling
         trigger: schedule
```
AnthonyHorton, line 73: Missing a transition from `ready` direct to `coarse_focusing`, for when we start up after it's too dark for evening flat fielding?

huntsman/observatory.py
```diff
@@ -195,6 +200,27 @@ def analyze_recent(self):
...
+        return False
+
+    def autofocus_cameras(self, *args, **kwargs):
+        '''
+        Override autofocus_cameras to update the last focus time.
+        '''
+        result = super().autofocus_cameras(*args, **kwargs)
+
+        # Update last focus time
+        self.last_focus_time = utils.current_time()
```
AnthonyHorton: I'd be slightly stricter and insist on there having been a fine focus recently, not either coarse or fine.

```suggestion
        coarse = kwargs.get("coarse", False)
        if not coarse:
            self.last_focus_time = utils.current_time()
```

AnthonyHorton (commented):
OK, this is, I think, an improvement. It's very hard to be sure though, because there are no unit tests for any of this! To be fair there have always been very limited tests for the Huntsman state machine, but we really need to fix that now. I think we're wasting a lot of time debugging state machine issues by running the full system on real hardware in real time that could be avoided with better unit tests.

huntsman/utils/states.py
```diff
@@ -0,0 +1,20 @@
...
+    # If the nearest midnight is in the past, its the morning...
+    return midnight < current_time()
+
+
+def wait_for_twilight(pocs):
+    '''
+    Wait for the Sun to be between flat and focus horizons.
+    '''
+    while pocs.is_dark(horizon='focus') or not pocs.is_dark(horizon='flat'):
+        pocs.sleep(delay=pocs._safe_delay)
```
wtgee: A `pocs` instance shouldn't really ever appear anywhere outside of the state files*.  Not that it necessarily couldn't, but it isn't how we have organized things. These should go alongside the other "Safety Methods" in `core.py`. We don't override anything in `core.py` for `huntsman-pocs` yet, but there's no reason not to.

(There's a case to be made for moving all of those methods out and doing them in smarter fashion, perhaps via some utils, but that's beyond the scope of this PR).

*We can also use a `pocs` instance in scripts, but that is different from it being in a `utils` module too.

wtgee (changes requested):

states/huntsman/coarse_focusing.py
AnthonyHorton: So, summary of the above we probably do want to remove the blanket catching of all exceptions here so that if an error occurs during autofocusing POCS will stop and we can work out what went wrong.

states/huntsman/sleeping.py
```diff
@@ -4,8 +4,10 @@ def on_enter(event_data):
     pocs.next_state = 'ready'
 
     # If it is dark and safe we shouldn't be in sleeping state
-    if pocs.is_dark() and pocs.is_safe():
-        pocs.say("Weather is good and it is dark. Something must have gone wrong. Stopping loop.")
+    if pocs.is_dark(horizon='observe') and pocs.is_safe():
```
AnthonyHorton: This is redundant in at least two ways, the `pocs.is_safe()` check includes checking whether it is dark and the default horizon for `pocs.is_dark()` is `observe` so we don't need to specify horizon unless we want to use one of the other. The equivalent line in the POCS version of `sleeping.py` is
`if pocs.is_safe() and pocs.should_retry is False:`

I'm also unsure that `observe` is the right horizon to use for Huntsman. Should Huntsman ever be sleeping when the Sun is below the `flat` horizon? If not then this check should be `if pocs.is_safe(horizon='flat'):` instead.

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+    return narrow_band_cameras, broad_band_cameras
+
+
+def _wait_for_twilight(pocs):
+    '''
+    Wait for twilight. Temporary solution until something better is found.
+    '''
+    pocs.logger.debug('Waiting for twilight...')
+    while pocs.is_dark(horizon='focus') or not pocs.is_safe(horizon='flat'):
+        pocs.sleep(delay=pocs._safe_delay)
```
AnthonyHorton: I'm a bit concerned that Huntsman could spend ~1/2 hour in this loop waiting for the Sun to rise from the `observe` horizon to the `focus` horizon and as far as I can tell there is no mechanism to break out of the loop and close the dome if the weather turns bad during that time. `pocs.sleep` includes calls to `pocs.check_messages()` every 10 seconds but I don't know if the weather system sends messages if the weather turns bad. @wtgee?

Adding
```
if not pocs.is_weather_safe():
    return
```
inside the `while` loop would ensure weather gets checked once every `._safe_delay` (which defaults to 5 minutes).

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+
+    return narrow_band_cameras, broad_band_cameras
+
+
+def _wait_for_twilight(pocs):
+    '''
+    Wait for twilight. Temporary solution until something better is found.
+    '''
+    pocs.logger.debug('Waiting for twilight...')
+    while pocs.is_dark(horizon='focus') or not pocs.is_safe(horizon='flat'):
```
AnthonyHorton: This combination of horizon and safety check is efficient but a bit opaque. To avoid future head scratching it should probably have a comment above it explaining what the intended condition we're waiting for actually is (i.e. Sun between 'focus' and 'flat' horizons, and safe).

AnthonyHorton: If the weather turns bad, or AC power fails, or the data disk fills up then `pocs.is_safe(horizon='flat')` could remain `False` for an arbitrarily long time. This could result in Huntsman getting stuck in this loop for days, or even weeks (if no one checks on it).

states/huntsman/ready.py
```diff
@@ -9,4 +9,20 @@ def on_enter(event_data):
...
+
+    # Don't need to focus, not dark enough to observe
+    elif pocs.observatory.past_midnight():
+        if pocs.is_dark(horizon='flat'):
+            pocs.next_state = 'twilight_flat_fielding'
+        else:
+            # We don't want to be waiting all day to take evening flats
+            pocs.next_state = 'parking'
+    else:
+        pocs.next_state = 'twilight_flat_fielding'
```
AnthonyHorton: You end up here any time a coarse focus is not needed, the Sun is above the `observe` horizon, and it is before midnight. The Sun could be below the `focus` horizon already, in which case the `twilight_flat_fielding` state will `_wait_for_twilight` all night long until it's ready to take morning flats.

states/huntsman/ready.py
```diff
@@ -9,4 +9,20 @@ def on_enter(event_data):
 
     pocs.observatory.mount.unpark()
```
AnthonyHorton, line 10: Good to set next state to parking right at the start as a fallback. Lots of states do this.

```suggestion
    pocs.next_state = 'parking'

    pocs.observatory.mount.unpark()
```

states/huntsman/ready.py
AnthonyHorton: This change fixes issue noted in comment above. Assumes `next_state` has already been set to `parking`.

```suggestion
    # Don't need to focus, not dark enough to observe
    else:
        if pocs.observatory.past_midnight():
            if pocs.is_dark(horizon='flat'):
                # After midnight and still dark enough to take flats
                pocs.next_state = 'twilight_flat_fielding'
        else:
            if not pocs.is_dark(horizon='focus'):
                # Before midnight and still light enough to take flats
                pocs.next_state = 'twilight_flat_fielding'
```

AnthonyHorton (changes requested):
Great to see a bunch more unit tests for the state machine! They're a good start, but the tests are still missing some of the less likely startup conditions that may catch us out (e.g. start up in the evening with Sun between `focus` and `observing` horizons but coarse focus not needed, I think I spotted a problem with that one). If anything it's more important for the unit tests to include the rare (but possible) situations than it is for them to include the usual situations because issues with normal operations will be found quickly when testing on the real hardware but bugs triggered by rare events could lurk for a long time.

Otherwise this is good, but I think I spotted a problem with the transition out of `ready` state (fix suggested) and also some ways in which the `twilight_flat_fields` state could get stuck waiting for twilight longer than it should. I'm also concerned about the safety of that wait for twilight, though I may be mistaken about that.

conf_files/huntsman.yaml
```diff
@@ -59,6 +59,11 @@ messaging:
     cmd_port: 6500
     msg_port: 6510
 state_machine: /var/huntsman/huntsman-pocs/resources/state_table/huntsman.yaml
+focusing:
+  coarse:
+    frequency: 2
+    frequency_unit: hour
```
wtgee, line 65: The `panoptes-utils` deserializer is already handling some astropy units, could be made to handle hour as well.

https://github.com/panoptes/panoptes-utils/blob/2cafd0d0863e434a6c698858461524b9bed6d949/panoptes/utils/serializers.py#L297

huntsman/observatory.py
```diff
@@ -104,6 +109,15 @@ def initialize(self):
             self.logger.debug("Connecting to autoguider")
             self.autoguider.connect()
 
+    def past_midnight(self):
+        """Check if its morning, useful for going into either morning or evening flats."""
```
```suggestion
        """Check if it's morning, useful for going into either morning or evening flats."""
```

huntsman/observatory.py
```diff
@@ -104,6 +109,15 @@ def initialize(self):
...
+
+        # Get the time of the nearest midnight to now
+        midnight = self.observer.midnight(utils.current_time(), which='nearest')
+
+        # If the nearest midnight is in the past, its the morning...
```
```suggestion
        # If the nearest midnight is in the past, it's the morning...
```

huntsman/observatory.py
```diff
@@ -104,6 +109,15 @@ def initialize(self):
             self.logger.debug("Connecting to autoguider")
             self.autoguider.connect()
 
+    def past_midnight(self):
+        """Check if its morning, useful for going into either morning or evening flats."""
+
+        # Get the time of the nearest midnight to now
+        midnight = self.observer.midnight(utils.current_time(), which='nearest')
```
wtgee: Standard usage is to import `current_time` directly so you don't need the `utils` out front. Same with  Not a big deal but I see a bunch of these here so thought worth mentioning.

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+# ==============================================================================
+
+
+def test_parking_ready(pocs):
+    '''
+    Test if the parking state transitions back into ready.
+    '''
+    pocs.initialize()
+    pocs.get_ready()
+    pocs.config['simulator'] = [_ for _ in pocs.config['simulator'] if _ != 'night']
```
wtgee: Using `_` is a bit odd to me. In Python `_` is almost universally used to indicate that the variable is very specifically **not** used. Here you are just using it as a temporary variable.

List comprehensions like this is where I think it is okay to use single letter variables if you feel like typing `simulator` (or `sim`) is too much:

```suggestion
    pocs.config['simulator'] = [s for s in pocs.config['simulator'] if s != 'night']
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+    os.environ['POCSTIME'] = '2016-08-13 23:00:00'
+    assert(not pocs.is_dark(horizon='observe'))
```
wtgee: Don't use paren calls on assert (lots of these in this file but should be easy to replace).

```suggestion
    assert not pocs.is_dark(horizon='observe')
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+    pocs.get_ready()
+    # Make sure its dark
+    assert(pocs.is_dark(horizon='observe'))
+    # Get into the sleeping state
+    pocs.next_state = 'parking'
+    for state in ['parking', 'parked', 'housekeeping', 'sleeping']:
+        assert(pocs.next_state == state)
+        if state == 'sleeping':
+            with pytest.raises(RuntimeError):
+                pocs.goto_next_state()
```
wtgee: Why does this throw a `RuntimeError` if it's dark?

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+    pocs.power_down()
+
+
+def test_evening_setup(pocs):
+    '''
+    Test the states in the evening before observing starts.
+    '''
+    # Preconditions to ready state
+    os.environ['POCSTIME'] = '2016-08-13 08:10:00'
+    pocs.config['simulator'] = [_ for _ in pocs.config['simulator'] if _ != 'night']
```
wtgee:
```suggestion
    pocs.config['simulator'] = [s for s in pocs.config['simulator'] if s != 'night']
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+        pocs.goto_next_state()
+    pocs.power_down()
+
+
+def test_evening_setup(pocs):
+    '''
+    Test the states in the evening before observing starts.
+    '''
+    # Preconditions to ready state
+    os.environ['POCSTIME'] = '2016-08-13 08:10:00'
```
wtgee: We might want to start picking more recent dates. The sunset calculation times take longer the further back you go.

I usually just pick whatever date I am writing the test on. @AnthonyHorton wanted to use 2020-02-02T20:20:02.02 as the base reference. ;)

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def test_morning_parking(pocs):
+    '''
+    Test morning transition between ready, flat fielding and parking.
+    '''
+    os.environ['POCSTIME'] = '2016-08-13 19:30:00'
+    pocs.initialize()
+    pocs.observatory.last_focus_time = utils.current_time()
+    pocs.config['simulator'] = [_ for _ in pocs.config['simulator'] if _ != 'night']
```
wtgee:
```suggestion
    pocs.config['simulator'] = [s for s in pocs.config['simulator'] if s != 'night']
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def test_morning_coarse_focusing_parking(pocs):
+    '''
+    Test morning transition between coarse focusing, flat fielding and parking.
+    '''
+    os.environ['POCSTIME'] = '2016-08-13 19:30:00'
+    pocs.initialize()
+    pocs.get_ready()
+    pocs.config['simulator'] = [_ for _ in pocs.config['simulator'] if _ != 'night']
```
wtgee:
```suggestion
    pocs.config['simulator'] = [s for s in pocs.config['simulator'] if s != 'night']
```

resources/state_table/huntsman.yaml
```diff
@@ -62,16 +65,28 @@ transitions:
...
+    -
+        source: coarse_focusing
+        dest: twilight_flat_fielding
+        trigger: take_twilight_flats
+    -
+        source: twilight_flat_fielding
+        dest: coarse_focusing
+        trigger: autofocus_coarse
```
wtgee, line 89: This needs to go both directions?

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # Morning and not dark enough for observing...
+    if pocs.observatory.past_midnight() and not pocs.is_dark(horizon='observe'):
```
wtgee: This is a very confusing way to ask "is it morning?" to me.

states/huntsman/sleeping.py
```diff
@@ -4,8 +4,10 @@ def on_enter(event_data):
...
         pocs.stop_states()
+        msg = "Weather is good and it is dark. Something must have gone wrong. Stopping loop."
+        pocs.say(msg)
+        raise RuntimeError(msg)
```
wtgee: Why the `RuntimeError` here?

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
+"""
+State to handle the taking of calibration frames (evening and morning).
+"""
+
+
+def _get_cameras(pocs):
```
wtgee: This and function below don't really need to be private functions.

Also, do we just want the `Observatory` itself to have this knowledge of camera types? @AnthonyHorton

states/huntsman/twilight_flat_fielding.py
wtgee: > I don't know if the weather system sends messages if the weather turns bad. @wtgee?

No, but ideally it should.

We use the `wait_until_safe` to handle just this kind of situation but it's not clear to me that will work here since what we want to wait on is a specific darkness level and not an overall safety.

@AnthonyHorton suggestion is okay but doesn't check things like AC and disk space.

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+
+    If evening, the next state will be coarse_focusing, else, parking.
+    '''
+    pocs = event_data.model
+    pocs.next_state = 'parking'
+
+    # Make sure its safe, dark and light enough for flats
+    _wait_for_twilight(pocs)
+
+    if pocs.observatory.take_flat_fields:
```
wtgee, line 52: Do we still even want this conditional? In what operational situations do we not want to take flats?

huntsman/observatory.py
```diff
@@ -104,6 +109,15 @@ def initialize(self):
             self.logger.debug("Connecting to autoguider")
             self.autoguider.connect()
 
+    def past_midnight(self):
```
wtgee: Based on how this is used below, this should be a property instead of a method.

Also, `past_midnight` seems a little ambiguous to me. I wonder if we could get more clarity with our conditionals by having a bunch of predicate properties: `is_evening`, `is_night`, `is_morning`, `is_day`.  We observe when it `is_night`, we calibrate when it `is_evening` or `is_morning`, we sleep when it `is_day`.

wtgee (commented):
In general looking good but I think we need to clean up a few things, as mentioned by @AnthonyHorton .

Also curious about my commet on `past_midnight` and if the observatory should have more knowledge of what time of day it is in easy property form.

states/huntsman/twilight_flat_fielding.py
danjampro: I think I had something similar in my original implementation. But yes, you are right!

AnthonyHorton: @wtgee Yep, some awkwardness comes from the fact that `is_safe()` is getting used for actual safety decisions (i.e. need to shut down because it's raining) and also for deciding whether it's the right time to do something (i.e. not too bright/not too dark, and also safe). Could do with either an additional method or an option for `is_safe()` so that we can easily just check weather, AC power and disk space independently of Sun altitude. Or maybe just use `is_safe()` as is, but with a horizon of 0 degrees Sun altitude, because running when the Sun is up is a genuine safety risk?

huntsman/tests/test_states.py
AnthonyHorton: Because of that Python convention the use of `_` here did briefly confuse me, before I realised that the values _weren't_ being thrown away....

states/huntsman/sleeping.py
danjampro: I think observe should still be used here. For example, if twilight flats finish in the morning then it is possible for Huntsman to end up sleeping while `is_dark(horizon='flat')`, but never while `is_dark(horizon='observe')`.

states/huntsman/twilight_flat_fielding.py
AnthonyHorton: @wtgee Yep, as I noted in the comment when I wrote this bit of code it's an ugly hack and was only ever intended to be a temporary solution. Now that we have filterwheels on Huntsman it's probably time to do this properly (in a different PR). We have 5 different filters available (luminance, g', r', H-alpha & [SII]), we'll want flats for all of them, they should have their flat fields taken in a particular order (reversed between evening/morning), and ideally flat fields for each filter should be taken within a particular narrow range of Sun altitudes.

So, we need more info in the config files, we need the `Observatory` to parse, store and make available that info, and we need a revamp of this state and the `Observatory` methods to correctly handle flat fielding a bunch of different filters (including changing filters).

states/huntsman/twilight_flat_fielding.py
AnthonyHorton: For Huntsman we'll always want to take them when we can, and since this is nominally Huntsman specific code then I guess it's not necessary.

states/huntsman/sleeping.py
AnthonyHorton: Good point. What about `is_safe(horizon='focus')` though? Morning twilights would not have happened yet so if it is safe and Sun is below the `focus` horizon Huntsman should not be sleeping.

states/huntsman/twilight_flat_fielding.py
danjampro: I think we can use `while pocs.is_safe(horizon='flat'):` here, because `pocs.is_dark(horizon='flat')` is required by the state machine to enter the state in the first place. Then we can just use `twilight = not pocs.is_dark(horizon='focus')` within the while loop.

resources/state_table/huntsman.yaml
danjampro: Yes, if we want to refocus in the morning

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,71 @@
...
+def wait_for_twilight(pocs):
+    '''
+    Wait for twilight. Temporary solution until something better is found.
+    '''
+    pocs.logger.debug('Waiting for twilight...')
+    while pocs.is_safe(horizon='flat'):
+        # Twilight when Sun between flat and focus horizons.
+        twilight = not pocs.is_dark(horizon='focus')
+        if not twilight:
+            pocs.sleep(delay=pocs._safe_delay)
```
AnthonyHorton: I don't think the single use, throwaway variable `twilight` is adding anything here, and I think it's actually clearer if you remove it, along with the double negative (two `not`).

```suggestion
        # If it's still too dark to take flats then wait some more.
        if pocs.is_dark(horizon='focus'):
            pocs.sleep(delay=pocs._safe_delay)
```

huntsman/tests/test_states.py
danjampro: I dont think it matters as long as it works. I mean, how long can it take??

states/huntsman/twilight_flat_fielding.py
danjampro: I included it because you said it wasn't clear enough that we were waiting for twilight.

huntsman/tests/test_states.py
danjampro: Because if its not supposed to be in the sleeping state then it should raise an error rather than continuing indefinitely which is what it currently does

huntsman/tests/test_states.py
danjampro: Why?

wtgee: It's not a function. https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

wtgee: Honestly I'm surprised it even works.

states/huntsman/twilight_flat_fielding.py
danjampro: Not necessary, no. But I've found it useful for unit testing the state machine without running all the code. Note that the `take_morning_flats` function is not yet implemented so the state machine tests would throw and error without this functionality. Its not perfect, but its not a problem for this PR to solve.

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    pocs.get_ready()
+    pocs.config['simulator'] = [s for s in pocs.config['simulator'] if s != 'night']
+    os.environ['POCSTIME'] = '2016-08-13 23:00:00'
+    assert not pocs.is_dark(horizon='observe')
+    pocs.next_state = 'parking'
+    for state in ['parking', 'parked', 'housekeeping', 'sleeping', 'ready']:
+        assert pocs.next_state == state
+        pocs.goto_next_state()
+        assert pocs.state == state
+    pocs.power_down()
```
wtgee: You do a `power_down` after the `yield` in the fixture, so this is not needed.
```suggestion
```

huntsman/tests/test_states.py
wtgee: It can take on the order of 60 seconds per calculation and it can run that calculation many times. It has been a preventative issue on a number of our PANOPTES units when the IERS tables cannot be updated. You can follow along with the issue here: https://github.com/astropy/astroplan/issues/441.

Also, because we end up with test failures because of precision errors in some of the tests (although that probably wouldn't happen here). Your two minutes of work saves lots of future confusion when tests start failing for no reason.

states/huntsman/coarse_focusing.py
```diff
@@ -0,0 +1,26 @@
+"""
+State to handle the coarse focusing at the start of the night.
```
wtgee:
```suggestion
State to handle the coarse focusing at the start of each scheduled observation.
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    assert not pocs.should_retry
+    pocs.next_state = 'parking'
+    for state in ['parking', 'parked', 'housekeeping', 'sleeping']:
+        assert pocs.next_state == state
+        if state == 'sleeping':
+            with pytest.raises(RuntimeError):
+                pocs.goto_next_state()
+        else:
+            pocs.goto_next_state()
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    Test if ready goes into observe if its dark enough.
+    '''
+    pocs.initialize()
+    pocs.observatory.last_focus_time = utils.current_time()
+    pocs.get_ready()
+    assert pocs.state == 'ready'
+    assert not pocs.observatory.require_coarse_focus()
+    assert pocs.is_dark(horizon='observe')
+    assert pocs.next_state == 'scheduling'
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    pocs.initialize()
+    pocs.observatory.last_focus_time = utils.current_time()
+    assert not pocs.observatory.past_midnight
+    assert not pocs.observatory.require_coarse_focus()
+    assert not pocs.is_dark(horizon='observe')
+    assert pocs.is_dark(horizon='focus')
+    pocs.get_ready()
+    assert pocs.state == 'ready'
+    assert pocs.next_state == 'scheduling'
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/observatory.py
```diff
@@ -195,6 +210,28 @@ def analyze_recent(self):
 
         return self.current_offset_info
 
+    def require_coarse_focus(self):
```
wtgee, line 213: This should be named something like `coarse_focus_required`. 

`require_coarse_focus` makes it sound like you are requiring a coarse focus, which is not what this method is doing.  The usage looks odd in some of the tests.

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    '''
+    pocs.initialize()
+    pocs.get_ready()
+    assert pocs.state == 'ready'
+    assert pocs.is_dark(horizon='observe')
+    assert pocs.observatory.require_coarse_focus()
+    for state in ['coarse_focusing', 'scheduling']:
+        assert pocs.next_state == state
+        pocs.goto_next_state()
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+        if state == 'twilight_flat_fielding':
+            assert not pocs.is_dark(horizon='focus')
+            assert pocs.is_dark(horizon='flat')
+        elif state == 'coarse_focusing':
+            os.environ['POCSTIME'] = '2016-08-13 08:40:00'
+            assert pocs.is_dark(horizon='focus')
+            assert not pocs.is_dark(horizon='observe')
+        pocs.goto_next_state()
+        assert pocs.state == state
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    assert not pocs.is_dark(horizon='observe')
+    assert pocs.is_dark(horizon='focus')
+    for state in ['twilight_flat_fielding', 'parking']:
+        assert pocs.next_state == state
+        if state == 'twilight_flat_fielding':
+            os.environ['POCSTIME'] = '2016-08-13 20:00:00'
+            assert pocs.is_dark(horizon='flat')
+        pocs.goto_next_state()
+        assert pocs.state == state
+    pocs.power_down()
```
wtgee:
```suggestion
```

huntsman/tests/test_states.py
```diff
@@ -0,0 +1,215 @@
...
+    assert not pocs.is_dark(horizon='observe')
+    assert pocs.is_dark(horizon='focus')
+    for state in ['coarse_focusing', 'twilight_flat_fielding', 'parking']:
+        assert pocs.next_state == state
+        if state == 'twilight_flat_fielding':
+            os.environ['POCSTIME'] = '2016-08-13 20:00:00'
+            assert pocs.is_dark(horizon='flat')
+        pocs.goto_next_state()
+        assert pocs.state == state
+    pocs.power_down()
```
wtgee:
```suggestion
```

states/huntsman/sleeping.py
```diff
@@ -4,8 +4,11 @@ def on_enter(event_data):
...
     # If it is dark and safe we shouldn't be in sleeping state
-    if pocs.is_dark() and pocs.is_safe():
-        pocs.say("Weather is good and it is dark. Something must have gone wrong. Stopping loop.")
-        pocs.stop_states()
+    if pocs.is_safe(horizon='focus'):
+        msg = "Weather is good and it is dark. Something must have gone wrong."
+        pocs.say(msg)
+        if not pocs.should_retry:
+            pocs.stop_states()
+            raise RuntimeError(msg)
```
wtgee:
```suggestion
    if pocs.is_safe(horizon='focus'):
        pocs.say("Weather is good and it is dark. Something must have gone wrong.")
        pocs.stop_states()
````


The `exit_when_done=True` param on the `POCS` constructor will cause the machine to exit gracefully after it has been stopped via `stop_states`. You probably just want to set that on your `pocs` instance. If we always want it to exit in that situation then can just change that default.

File: states/huntsman/twilight_flat_fielding.py [hunk 50]
```diff
@@ -0,0 +1,76 @@
...
+    '''
+    Calibrating state. If safe to do so, take flats and darks. Should be
+    called once at the beginning and end of the night.
+
+    If evening, the next state will be coarse_focusing, else, parking.
+    '''
+    pocs = event_data.model
+    pocs.next_state = 'parking'
+
+    # Make sure its safe, dark and light enough for flats
```
wtgee:
```suggestion
    # Make sure it's safe, dark and light enough for flats
```

huntsman/tests/test_states.py
AnthonyHorton: Might want to use one of the custom errors from POCS utils (or `panoptes-utils`) in that case, e.g. `PanError`.

resources/state_table/huntsman.yaml
AnthonyHorton: There might be an argument for separate morning and evening twilight flat states to disentangle the state machine diagram a bit. Each state would only contain the code needed to customise things for evening or morning, plus calls to generic methods of `Observatory` or `POCS`.

AnthonyHorton (approved):
OK, I think I'm done nitpicking now. LGTM (subject to any changes requested by @wtgee)!

states/huntsman/twilight_flat_fielding.py
```diff
@@ -0,0 +1,76 @@
...
+
+def wait_for_twilight(pocs):
+    '''
+    Wait for twilight. Temporary solution until something better is found.
+
+    Twilight when Sun between flat and focus horizons.
+    '''
+    pocs.logger.debug('Waiting for twilight...')
+    while pocs.is_safe(horizon='flat'):
+        if pocs.is_dark(horizon='focus'):
```
wtgee, line 31: It still seems like if you start observing and get into this state after the sun has passed `focus` that this will then wait until the morning.

huntsman/tests/test_states.py
wtgee: See note below.

wtgee (approved):
Mostly nitpicking. :wink:

states/huntsman/twilight_flat_fielding.py
danjampro: Yes it will - and that's sort of desirable behaviour in the morning. At the moment the logic for entering the state should prevent it entering too early.

PR #181, opened 2020-04-19
Improve docker images 

danjampro:
This PR improves on the existing Huntsman camera docker image by:

a) Making it 0.5 Gb smaller 
b) Making an abstract arm_device base image that can also be used for e.g. domepi
c) Adding passwordless sudo rights to Huntsman user
d) Adding docker-compose support
e) Mounting the camera logs outside of the docker container

Tested on pi4 and camera server started as normal.


fergusL (approved):
LGTM!

docker/arm_device/dockerfile
```diff
@@ -0,0 +1,66 @@
+#==============================================================================
```
wtgee, line 1: Should change the filename to `Dockerfile`, which is what is expected by all automatic services.

docker/arm_device/dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+USER root
+RUN ln -s $CONDA_PATH/bin/conda /usr/local/bin/ && \
+    ln -s $CONDA_PATH/bin/python3.6 /usr/local/bin/python && \
+    ln -s $CONDA_PATH/bin/pip /usr/local/bin/ && \
+    # Install python deps
+    pip install --upgrade pip && \
+    pip install setuptools --upgrade && \
+    conda install astropy matplotlib requests scipy && \
+    pip install netifaces pyro4 pyyaml pyserial pymongo && \
```
wtgee:
```suggestion
    pip install netifaces pyro4 pyyaml pyserial && \
```

docker/arm_device/startup.sh
```diff
@@ -0,0 +1,4 @@
+sudo chown -R $PANUSER /home/$PANUSER/.ssh
```
wtgee, line 2: Add a `#!/bin/sh` shebang.

wtgee (approved):
Looks good. If we are worried about image size at this point there is still a bit that could be done to reduce it, but I'm certainly not an expert on that.  And I don't think it's the primary concern right now.

PR #182, opened 2020-04-27
Dockerised Huntsman testing environment

danjampro:
This PR creates a docker image specifically for testing new huntsman-pocs code. With a single command, you can fire up a container which will automatically run the specified tests and display the logs in a Byobu window, using a local version of huntsman-pocs located at `$HUNTSMAN_POCS` on the local system.

Eventually, this image should use the Huntsman control computer docker image as its base. I'll add to the wiki once merged.


docker/testenv/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+# Install required packages
+RUN apt-get update && \
+    apt-get -y install build-essential && \
+    apt-get -y install byobu sshfs vim grc sudo && \
+    apt install -y openssh-server --fix-missing && \
+    apt install -y openssh-client && \
+
+    # Setup PANUSER
+    useradd -m -d /home/$PANUSER $PANUSER && \
+    echo $PANUSER:$PANUSER | chpasswd && \
```
wtgee: What's going on here?  The `useradd` function should be able to set a default password.

wtgee (changes requested):
The new Dockerfile largely recreates everything that exists in the main Dockerfiles. We should be trying to layer these instead.

I have done largely this exact same thing within `panoptes-utils`.  Check out the [panoptes-utils/docker](https://github.com/panoptes/panoptes-utils/tree/develop/docker) directory (although ignore the README, I'm realizing it's outdated).

Specifically, I'm just creating the same image as normal except with a `develop` tag instead of `latest`.  This image is built on top of `latest` and just adds needed development and testing things.  

I would say should have something similar to that rather than trying to maintain multiple almost identical Dockerfiles.

danjampro:
> The new Dockerfile largely recreates everything that exists in the main Dockerfiles. We should be trying to layer these instead.
> 
> I have done largely this exact same thing within `panoptes-utils`. Check out the [panoptes-utils/docker](https://github.com/panoptes/panoptes-utils/tree/develop/docker) directory (although ignore the README, I'm realizing it's outdated).
> 
> Specifically, I'm just creating the same image as normal except with a `develop` tag instead of `latest`. This image is built on top of `latest` and just adds needed development and testing things.
> 
> I would say should have something similar to that rather than trying to maintain multiple almost identical Dockerfiles.

Hi @wtgee thanks for the review. While I certainly agree with what you're saying, I don't think I will get around to this for a while - creating a docker image for the control computer has been on the agenda for some time but is not currently high priority. 

Having said that, I do think this functionality is useful (if not for anyone else, for me) for testing code now and in the near future, so I will merge it but make an issue that we should make efforts to layer the docker images in future.

PR #184, opened 2020-05-17
Ready cameras

danjampro:
Small PR to make sure all cameras are ready before leaving the `ready` state. Also deactivates camera cooling in `housekeeping` state.


states/huntsman/housekeeping.py
```diff
@@ -1,5 +1,17 @@
+def deactivate_camera_cooling(pocs):
+    """
+    Deactivate camera cooling before entering sleeping state.
+    """
+    pocs.logger.debug('Deactivating camera cooling before entering sleeping state.')
+    for cam in pocs.observatory.cameras.values():
+        if cam.is_cooled_camera:
+            cam.cooling_enabled = False
```
wtgee: A couple of things here, although some may be a different PR.

Namely, setting `cooling_enabled` should handle the `if is_cooled_camera` so that we don't have to check for it before setting. I thought this was already the case but if not we should make it the case (perhaps the separate PR).

Assuming you don't have the conditional check, this really doesn't need to be a separate function. Even with the conditional there is not a ton of need for it.

Alternatively, if "deactivating all the cooling" is something we find useful, this should move into `POCS` or `Observatory` as a public method.

states/huntsman/ready.py
```diff
@@ -1,3 +1,27 @@
...
+    """
+    Make sure cameras are ready before starting states.
+    """
+    pocs.logger.debug('Preparing cameras before leaving ready state.')
+    cameras = pocs.observatory.cameras.values()
+
+    # Make sure camera cooling is enabled
+    for cam in cameras:
+        if cam.is_cooled_camera:
+            cam.cooling_enabled = True
```
wtgee: Same as comment above.

states/huntsman/ready.py
```diff
@@ -1,3 +1,27 @@
...
+
+    # Wait for cameras to be ready
+    for i in range(max_attempts):
+        if all([cam.is_ready() for cam in cameras]):
+            return
+        time.sleep(sleep)
+    raise error.PanError('Timeout while waiting for cameras to become ready from ready state.')
```
wtgee: I don't think we should stop all execution if a single camera is not working.  PANOPTES currently does this with just two cameras and it causes a lot of downtime.

If we are dynamically removing cameras from `pocs.observatory.cameras` when they enter an error state than this might be okay.

states/huntsman/ready.py
```diff
@@ -32,3 +56,7 @@ def on_enter(event_data):
                 pocs.next_state = 'scheduling'
             else:
                 pocs.next_state = 'twilight_flat_fielding'
+
+    # Prepare the cameras if we are about to take some exposures
+    if pocs.next_state != 'parking':
+        prepare_cameras(pocs)
```
wtgee: This also probably doesn't need to be a separate function given the above comments.

wtgee (commented):
Leaving some comments for feedback while waiting for tests.

states/huntsman/ready.py
danjampro: Not sure I agree with that one. Until we have some way to notify us of any errors (e.g. cameras not working) I would rather an error be raised immediately so we realise what's happened.

states/huntsman/housekeeping.py
danjampro: I'm not sure whether `cooling_enabled` is handled like that - I can't see it in `POCS`. Maybe @AnthonyHorton can clarify? 

Maybe there isn't much need to have it as a separate function, but I like the idea of keeping the state codes as abstract as possible. 

Not sure moving the function to `observatory` is really necessary at the moment given its only going to be called here, unless cooling is turned off somewhere else?

danjampro: On second thought, maybe it would be better to add `deactivate_camera_cooling` and `activate_camera_cooling` methods to `HuntsmanObservatory`.

AnthonyHorton: Currently if you try to set `cooling_enabled` on a camera that doesn't implement that control it will raise a `NotImplementedError`. So, either need to check `is_cooled_camera` before setting it, or wrap the setting in a `try ... except NotImplementError` or, perhaps best of all, wrap it in `with suppress(NotImplementedError):`.

@danjampro this is all in the `pocs.camera.camera.AbstractCamera` base class.

states/huntsman/ready.py
AnthonyHorton: You could argue that one the other way around. If we don't have some way to notify us of errors then it's even more important for Huntsman to just deactivate a problematic camera and carries on observing without it, because the alternative is Huntsman not doing any more observing until someone checks on it.

For errors that don't compromise safety (e.g. camera or focuser errors) I do think that Huntsman should deactivate the effected system and carry on observing if possible.

danjampro: I agree that might be a reasonable approach during normal observing. But we are still in the early phases of commissioning and I would like keep hardware errors as self-evident as possible over this time. Maybe it could be implemented at a later phase in commissioning?

huntsman/observatory.py
```diff
@@ -316,6 +316,43 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+        Deactivate camera cooling for all cameras.
+        """
+        self.logger.debug('Deactivating camera cooling for all cameras.')
+        for cam in self.cameras.values():
+            if cam.is_cooled_camera:
+                cam.cooling_enabled = False
+
+    def prepare_cameras(self, sleep=60, max_attempts=5):
+        """
+        Make sure cameras are all cooled and ready.
```
wtgee, line 339:
```suggestion
        Make sure cameras are cooled and ready.
```

huntsman/observatory.py
```diff
@@ -316,6 +316,43 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                cam.cooling_enabled = True
+
+    def deactivate_camera_cooling(self):
+        """
+        Deactivate camera cooling for all cameras.
+        """
+        self.logger.debug('Deactivating camera cooling for all cameras.')
+        for cam in self.cameras.values():
+            if cam.is_cooled_camera:
+                cam.cooling_enabled = False
```
wtgee, line 335: Part of me would argue for a `toggle_camera_cooling` instead, but I realize you really, really want to be done with this PR. :grin:

huntsman/observatory.py
```diff
@@ -316,6 +316,43 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+        # Make sure camera cooling is enabled
+        self.activate_camera_cooling()
+
+        # Wait for cameras to be ready
+        self.logger.debug('Waiting for cameras to be ready.')
+        for i in range(max_attempts):
+            if all([cam.is_ready for cam in self.cameras.values()]):
+                return
+            time.sleep(sleep)
+        raise error.PanError('Timeout while waiting for cameras to be ready.')
```
wtgee: Ideally this is where we would have the boolean option `require_all_cameras`.  In a first pass this could be done with the for loop, it just means that you would always hit the `max_attempts` if any of the cameras were taking a long time to start.  

That should then be switched somehow so the camera returns an event and this implements some other loop to wait and check.

Anyway, just marking for a future PR.

states/huntsman/ready.py
```diff
@@ -32,3 +32,7 @@ def on_enter(event_data):
                 pocs.next_state = 'scheduling'
             else:
                 pocs.next_state = 'twilight_flat_fielding'
+
+    # Prepare the cameras if we are about to take some exposures
+    if pocs.next_state != 'parking':
+        pocs.observatory.prepare_cameras()
```
wtgee:
```suggestion
    if pocs.next_state != 'parking':
        pocs.say("Those cameras are hot! Let's turn on the cooling. This might take a few minutes.")
        pocs.observatory.prepare_cameras()
```

states/huntsman/housekeeping.py
```diff
@@ -11,4 +13,7 @@ def on_enter(event_data):
     except Exception as e:
         pocs.logger.warning('Problem with cleanup: {}'.format(e))
 
+    # Turn-off camera cooling
+    pocs.observatory.deactivate_camera_cooling()
```
wtgee, line 18:
```suggestion
    pocs.say('Shutting down for the night, going to turn off the camera cooling')
    pocs.observatory.deactivate_camera_cooling()
```

huntsman/observatory.py
```diff
@@ -316,6 +316,43 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+        """
+        # Make sure camera cooling is enabled
+        self.activate_camera_cooling()
+
+        # Wait for cameras to be ready
+        self.logger.debug('Waiting for cameras to be ready.')
+        for i in range(max_attempts):
+            if all([cam.is_ready for cam in self.cameras.values()]):
+                return
+            time.sleep(sleep)
```
```suggestion
        for i in range(max_attempts):
            pocs.logger.debug(f'Preparing cameras {i}/{max_attempts}')                        
            if all([cam.is_ready for cam in self.cameras.values()]):
                return
            pocs.say('Not all the cameras are cooling yet, going to give them a little more time')
            time.sleep(sleep)
```

wtgee (approved):
Thanks for all the work on this! I mostly added some logging calls.

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                    msg = f'Timeout while waiting for {cam_name} to be ready.'
+
+                    # Raise PanError if we need all cameras
+                    if require_all_cameras:
+                        raise error.PanError(msg)
+
+                    # Drop the camera if we don't need all cameras
+                    else:
+                        self.logger.error(msg)
+                        self.logger.debug(f'Removing {cam_name} from {self}.')
```
wtgee, line 374: `remove_camera` already logs the fact that it's being removed.  For extra logging here we should specify where it is coming from, something like:

```suggestion
                        self.logger.debug(f'{cam_name} not ready after {max_attempts*sleep} seconds, removing from {self}.')
```

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                If `True` and max_attempts is reached, a `PanError` will be raised. If `False`,
+                any camera that has failed to become ready will be dropped from the Observatory.
+        """
+        # Make sure camera cooling is enabled
+        self.activate_camera_cooling()
+
+        # Wait for cameras to be ready
+        n_cameras = len(self.cameras)
+        self.logger.debug('Waiting for cameras to be ready.')
+        for i in range(max_attempts):
```
wtgee: Easier and more reliable to handle the subtraction here:
```suggestion
        for i in range(max_attempts - 1):
```

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+
+            num_cameras_ready = 0
+            for cam_name, cam in self.cameras.items():
+
+                if cam.is_ready:
+                    num_cameras_ready += 1
+                    continue
+
+                # If max attempts have been reached...
+                if (i == max_attempts-1):
```
```suggestion
                if (i == max_attempts):
```

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                        self.logger.debug(f'Removing {cam_name} from {self}.')
+                        self.remove_camera(cam_name)
+
+            # Terminate loop if all cameras are ready
+            self.logger.debug(f'Number of ready cameras after {i+1} of {max_attempts} checks:'
+                              f' {num_cameras_ready} of {n_cameras}.')
+            if num_cameras_ready == n_cameras:
+                self.logger.debug('All cameras are ready.')
+                break
+            elif i < max_attempts-1:
```
```suggestion
            elif i < max_attempts:
```

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                        raise error.PanError(msg)
+
+                    # Drop the camera if we don't need all cameras
+                    else:
+                        self.logger.error(msg)
+                        self.logger.debug(f'Removing {cam_name} from {self}.')
+                        self.remove_camera(cam_name)
+
+            # Terminate loop if all cameras are ready
+            self.logger.debug(f'Number of ready cameras after {i+1} of {max_attempts} checks:'
```
```suggestion
            self.logger.debug(f'Number of ready cameras after {i+1} of {max_attempts+1} checks:'
```

huntsman/observatory.py
```diff
@@ -316,6 +316,79 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
...
+                self.logger.debug('All cameras are ready.')
+                break
+            elif i < max_attempts-1:
+                self.logger.debug('Not all cameras are ready yet, '
+                                  f'waiting another {sleep} seconds before checking again.')
+                time.sleep(sleep)
+
+        # Raise a `PanError` if no cameras are ready.
+        if num_cameras_ready == 0:
+            raise error.PanError('No cameras ready after maximum attempts reached.')
```
wtgee, line 390: `huntsman-pocs` should start creating it's own custom error classes and using those. We don't quite utilize exceptions efficiently now, but ideally this would raise something like `error.CameraNotReady`.

Future PR.

wtgee (approved):
:+1:

huntsman/observatory.py
danjampro: Is that right? for `max_attempts=3`, i will be (0, 1, 2).

huntsman/observatory.py
danjampro: This will cause the system to sleep after the last iteration, which is unnecessary.

huntsman/observatory.py
danjampro: Doesn't this just make the number of checks decrease by 1? Perhaps you mean `for i in range(1, max_attempts+1)`?

PR #186, opened 2020-07-14
Update download links for ZWO libraries

danjampro:
Tested links work.


danjampro:
Merging because minor change.

PR #187, opened 2020-07-15
Periodically update config file 

danjampro:
Simple change to avoid having to reset config server each if we change the config file.

### DIFF
--- a/huntsman/utils/config.py
+++ b/huntsman/utils/config.py
@@ -1,40 +1,48 @@
-#!/usr/bin/env python3
-# -*- coding: utf-8 -*-
-"""
-Created on Thu Jan  2 12:04:11 2020
-
-@author: danjampro
-
-Code to provide a config server using pyro.
-"""
 import sys
 import time
 import Pyro4
-from huntsman.utils import load_config, get_own_ip, DummyLogger
+from pocs.utils import current_time
 from pocs.utils.config import _parse_config
+from huntsman.utils import load_config, get_own_ip, DummyLogger
 
 
 @Pyro4.expose
 class ConfigServer():
-    '''
-    Class representing the config server.
-    '''
 
-    def __init__(self, config_file=None, parse=True, **kwargs):
-        '''
-
-        '''
-        self.parse = parse
+    def __init__(self, config_file=None, parse=True, refresh_frequency=120,
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        config_file (str):
+            The name of the config file to use. Default is device_info.yaml.
+        parse (bool):
+            Parse the config? Default True.
+        refresh_frequency (float):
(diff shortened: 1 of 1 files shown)


huntsman/utils/config.py
```diff
@@ -1,58 +1,66 @@
...
+    def __init__(self, config_file=None, parse=True, refresh_frequency=120,
+                 **kwargs):
+        """
+        Parameters
+        ----------
+        config_file (str):
+            The name of the config file to use. Default is device_info.yaml.
+        parse (bool):
+            Parse the config? Default True.
+        refresh_frequency (float):
```
AnthonyHorton: This is a refresh interval, not a refresh frequency.

AnthonyHorton (commented):
I can see why this could be handy, but in order to make most changes to the config stick we would still need to restart all the camera servers and/or POCS anyway so by itself it doesn't achieve much.

Is the idea that this would be the first step in a broader set of changes that would also, for example, automatically restart a camera sever when its section of the config changes? If so there should be an overall Issue with a checklist to keep track of the various changes required to achieve the end goal.

danjampro:
> I can see why this could be handy, but in order to make most changes to the config stick we would still need to restart all the camera servers and/or POCS anyway so by itself it doesn't achieve much.
> 
> Is the idea that this would be the first step in a broader set of changes that would also, for example, automatically restart a camera sever when its section of the config changes? If so there should be an overall Issue with a checklist to keep track of the various changes required to achieve the end goal.

@AnthonyHorton , thanks for taking a look. I don't have much of a grand plan for this at the moment - it is more of a quick modification for (mainly my own) convenience during commissioning. The overall plan is something we need to discus with @wtgee, as we might end up using the `panoptes-utils` version of the config server anyway.

wtgee:
> > I can see why this could be handy, but in order to make most changes to the config stick we would still need to restart all the camera servers and/or POCS anyway so by itself it doesn't achieve much.
> > Is the idea that this would be the first step in a broader set of changes that would also, for example, automatically restart a camera sever when its section of the config changes? If so there should be an overall Issue with a checklist to keep track of the various changes required to achieve the end goal.
> 
> @AnthonyHorton , thanks for taking a look. I don't have much of a grand plan for this at the moment - it is more of a quick modification for (mainly my own) convenience during commissioning. The overall plan is something we need to discus with @wtgee, as we might end up using the `panoptes-utils` version of the config server anyway.

Yes, I was going to make some comments about the grand structure, but that's what I had wanted to talk about in person (sorry for missing meeting!).  I can give a quick review with the idea that we will be revisiting the config server in general (for instance, I realized with this PR that in `huntsman-pocs` the config server is not really being used as a server).

wtgee (approved):
As mentioned above, I'm just going to approve if it's helping you develop or during testing runs. Let's plan to revisit config as a bigger topic soon.

PR #188, opened 2020-07-15
Improve prepare cameras

danjampro:
We now require several calls to `camera.is_ready` to return `True` consecutively in order for a camera to be considered as ready in `prepare_cameras`. This is to ensure the temperature has stabilised after cooling has been enabled.

The new argument `consecutive` specifies how many consecutive successes are required.

### DIFF
--- a/huntsman/observatory.py
+++ b/huntsman/observatory.py
@@ -425,47 +425,52 @@ class HuntsmanObservatory(Observatory):
             if cam.is_cooled_camera:
                 cam.cooling_enabled = False
 
-    def prepare_cameras(self, sleep=60, max_attempts=5, require_all_cameras=False):
+    def prepare_cameras(self, sleep=60, max_attempts=10, consecutive=3,
+                        require_all_cameras=False):
         """
         Make sure cameras are all cooled and ready.
 
         Arguments:
             sleep (float): Time in seconds to sleep between checking readiness. Default 60.
+            consecutive (int): Number of consecutive readies to be considered ready.
             max_attempts (int): Maximum number of ready checks. See `require_all_cameras`.
             require_all_cameras (bool): `True` if all cameras are required to be ready.
                 If `True` and max_attempts is reached, a `PanError` will be raised. If `False`,
                 any camera that has failed to become ready will be dropped from the Observatory.
         """
+        # Consecutive needs to be less than or equal to max_attempts
+        if consecutive > max_attempts:
+            max_attempts = consecutive
+
         # Make sure camera cooling is enabled
         self.activate_camera_cooling()
 
-        # Wait for cameras to be ready
+        # Setup containers
         n_cameras = len(self.cameras)
+        cameras_failed = []
+        ready_counts = {cam_name: 0 for cam_name in self.cameras.keys()}
+
+        # Wait for cameras to be ready
         self.logger.debug('Waiting for cameras to be ready.')
         for i in range(1, max_attempts+1):
 
             num_cameras_ready = 0
             for cam_name, cam in self.cameras.items():
 
+                # Check if camera is ready
                 if cam.is_ready:
-                    num_cameras_ready += 1
-                    continue
+                    ready_counts[cam_name] += 1
+
+                else:
+                    # Reset ready counts for this camera
+                    ready_counts[cam_name] = 0
 
                 # If max attempts have been reached...
-                if (i == max_attempts):
-                    msg = f'Timeout while waiting for {cam_name} to be ready.'
+                if (i == max_attempts) & (ready_counts[cam_name] < consecutive):
+                    cameras_failed.append(cam_name)
 
-                    # Raise PanError if we need all cameras
-                    if require_all_cameras:
-                        raise error.PanError(msg)
-
-                    # Drop the camera if we don't need all cameras
-                    else:
-                        self.logger.error(msg)
-                        self.logger.debug(f'Removing {cam_name} from {self}.')
-                        self.remove_camera(cam_name)
-
-            # Terminate loop if all cameras are ready
+            # Check if all cameras are ready
+            num_cameras_ready = sum([ready_counts[c] >= consecutive for c in self.cameras.keys()])
             self.logger.debug(f'Number of ready cameras after {i} of {max_attempts} checks:'
                               f' {num_cameras_ready} of {n_cameras}.')
             if num_cameras_ready == n_cameras:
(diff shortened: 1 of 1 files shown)


huntsman/observatory.py
```diff
@@ -425,47 +425,52 @@ def deactivate_camera_cooling(self):
...
             sleep (float): Time in seconds to sleep between checking readiness. Default 60.
+            consecutive (int): Number of consecutive readies to be considered ready.
             max_attempts (int): Maximum number of ready checks. See `require_all_cameras`.
             require_all_cameras (bool): `True` if all cameras are required to be ready.
                 If `True` and max_attempts is reached, a `PanError` will be raised. If `False`,
                 any camera that has failed to become ready will be dropped from the Observatory.
         """
+        # Consecutive needs to be less than or equal to max_attempts
+        if consecutive > max_attempts:
+            max_attempts = consecutive
```
AnthonyHorton, line 443: I don't think it's good practice to silently fix an invalid argument value like this. I'd be more inclined to just raise a descriptive `ValueError` in this sort of situation, and if you are going to override the argument value instead then at the very least you should log a warning, and probably display it too (using `warnings.warn`).

huntsman/observatory.py
```diff
@@ -476,6 +481,22 @@ def prepare_cameras(self, sleep=60, max_attempts=5, require_all_cameras=False):
                                   f'waiting another {sleep} seconds before checking again.')
                 time.sleep(sleep)
```
AnthonyHorton, line 482: A more idiomatic alternative implementation:

```
ready_counts = {cam_name: 0 for cam_name in self.cameras.keys()}

for i in range(max_attempts):
    for cam_name, cam in self.cameras.items():
        if cam.is_ready:
            ready_counts[cam_name] += 1
        else:
            ready_counts[cam_name] = 0

    if all([count >= consecutive for count in ready_counts.values()]):
        # All cameras are ready so stop checking now.
        break

    time.sleep(sleep)

failed_cameras = [cam_name for cam_name, count in ready_count.items() if count < consecutive]
if failed_cameras:
    # Handle failed cameras
else:
    # All cameras ready
```

Simplifies things a bit, removing a redundant check of the loop index, replacing some explicit loops with list comprehensions, and removing some unnecessary temporary variables

huntsman/observatory.py
```diff
@@ -476,6 +481,22 @@ def prepare_cameras(self, sleep=60, max_attempts=5, require_all_cameras=False):
...
 
+        # Handle failed cameras
+        if len(cameras_failed) != 0:
+            msg = "The following cameras have failed to become ready: "
+            msg += ", ".join(cameras_failed)
```
AnthonyHorton, line 487: Might be useful at this point to log why the camera is not ready (temperature unstable, exposing, filterwheel or focuser not ready).

AnthonyHorton (changes requested):
Looks good, though the main loop can be cleaned up a bit, and the input validation and error logging could be improved.

huntsman/observatory.py
```diff
@@ -476,6 +481,22 @@ def prepare_cameras(self, sleep=60, max_attempts=5, require_all_cameras=False):
...
+
+            # Raise PanError if we need all cameras
+            if require_all_cameras:
+                raise error.PanError(msg)
+
+            # Drop the camera if we don't need all cameras
+            else:
+                msg += ". Dropping them from the Observatory."
```
wtgee, line 495: The continued `msg` is a bit weird to me. I would just have the message in the `PanError` above and not have this part of the message or have it as a separate log call.

wtgee (commented):
In general I'm confused as to why we need to ask the cameras if they are ready multiple times.

Seems like this should be restructured so that the method is built around how much time it wants to wait while the cameras get ready, namely a `timeout` param.

Here loosely implemented with CountdownTimer:

(`unready_cameras` is a bad name but I need more :coffee:)
```py
timer = CountdownTimer(300)  # 5 minutes
sleep_interval = 60 # Should really be a part of CountdownTimer

unready_cameras = set(self.cameras.keys())
ready_cameras = set()

while timer.expired() is False:
    for cam_name in unready_cameras():
        if self.cameras[cam_name].is_ready():
            ready_cameras.add(cam_name)
            unready_cameras.remove(cam_name)

    timer.sleep(max_sleep=sleep_interval)

# Deal with ready and unready cameras below
```

I guess my real question is why do the cameras report ready and then not ready?  Can we implement a `stable_time` or something to the cameras such that they will only report ready if they have been (and stayed=) cooled for a certain amount of time?

AnthonyHorton:
Yeah, there really ought to be an Issue that explains the problem that this PR is intended to solve. This also gives you an extra feeling of achievement when you merge a PR and close an Issue at the same time...

@wtgee In short, the motivation for this is that we have observed that the ZWO cameras have a tendency to overshoot the target temperature when cooling. Consequently we ran into confusing errors where all the cameras would report ready, then when POCS moved on to taking exposures we would get errors about unstable temperatures because some of the cameras had gone straight through the target temperature and were outside the temperature tolerance on the low side. So, the aim here is to wait until the cameras are consistently reporting ready before moving on. I agree that this would be better expressed as a minimum time that they should report ready for, rather than the number of consecutive queries.

wtgee:
On Sat, Jul 18, 2020, 13:28 Anthony Horton <notifications@github.com> wrote:

> Yeah, there really ought to be an Issue that explains the problem that
> this PR is intended to solve. This also gives you an extra feeling of
> achievement when you merge a PR and close an Issue at the same time...
>
> @wtgee <https://github.com/wtgee> In short, the motivation for this is
> that we have observed that the ZWO cameras have a tendency to overshoot the
> target temperature when cooling. Consequently we ran into confusing errors
> where all the cameras would report ready, then when POCS moved on to taking
> exposures we would get errors about unstable temperatures because some of
> the cameras had gone straight through the target temperature and were
> outside the temperature tolerance on the low side. So, the aim here is to
> wait until the cameras are consistently reporting ready before moving on. I
> agree that this would be better expressed as a minimum time that they
> should report ready for, rather than the number of consecutive queries.
>
But can't we just build some kind of `settle_time` into the `is_ready`
call?

This just feels a bit hacky. If we are asking a camera if it is ready and
is reporting yes but that's not true (because it hasn't settled), then I
would prefer to fix the `is_ready` so that it's always accurate.

This would be in addition to changing the logic here to be time based.

AnthonyHorton:
> But can't we just build some kind of `settle_time` into the `is_ready` call? This just feels a bit hacky. If we are asking a camera if it is ready and is reporting yes but that's not true (because it hasn't settled), then I would prefer to fix the `is_ready` so that it's always accurate. This would be in addition to changing the logic here to be time based. —

Yes, but it gets a bit complicated because we can't afford for every `is_ready` check to take several minutes while it confirms that the temperature of each camera is stable, in turn. `is_ready` gets accessed at the start of every exposure.

We'd need to do something like having the cameras spawn their own temperature monitoring threads to keep track of the recent temperature history so that when we access `is_temperature_stable` it can immediately confirm whether the temperature has consistently stayed within tolerance of the target temperature for the preceding n minutes.

This level of sophistication is only actually required immediately after turning the cooling on, or making a large change to the target temperature.

What Dan's doing to achieve the same effect in a method of `Observatory`instead is undeniably a bit a hack, but on the other hand it means we're only doing it when it is most needed (just after turning cooling on) and it makes it easy to do the waiting for each camera in parallel without spawning lots of threads.

danjampro:
Thanks for the comments @AnthonyHorton @wtgee. I agree the time-based approach is more succinct.

I also like @AnthonyHorton's suggestion of having cameras keep track of their most recent temperature history, and actually prefer it to our current approach. It sounds like a PR for `POCS` though. Does that also sound like a good idea to you @wtgee ?

PR #190, opened 2020-08-14
Restructure into src directory

danjampro:
Moves python files into `src/huntsman/pocs`. This is consistent with the recent changes to `panoptes` repositories. Huntsman code is now imported like `from huntsman.pocs.x import yy` rather than `from huntsman.x import yy`, which will allow us to import code from `huntsman-drp` in a similar fashion (i.e. under the huntsman namespace). 

@wtgee will expand on these changes when he implements `pyscafold`.


setup.cfg
```diff
@@ -19,6 +19,13 @@ all_files = 1
 upload-dir = docs/_build/html
 show-response = 1
 
+[options]
+zip_safe = False
+packages = find:
+include_package_data = True
+package_dir =
+    =src
+
```
wtgee, line 28: Have you attempted installing this yet? Either via `pip` or `python setup.py`.

I'm mostly wondering if you have tested the namespace and seen if you get conflicts or import errors.  `pyscaffold` does a few other things to make this work IIRC.

setup.py
```diff
@@ -30,7 +28,7 @@
 #            if os.path.basename(fname) != 'README.rst']
 
 setup(name=PACKAGENAME,
-      version=__version__,
+      version="0.1.1",
```
wtgee: You didn't delete the version file so just curious why this hard-coding?

This probably breaks the regular `__version__` for now, but I'm sure that's fine.

wtgee (approved):
The ones here look good. The issue would of course be any missed ones that aren't here. But I'm guessing this is a basic search and replace and anything else will (hopefully!) show up in tests.

So I think for the namespace to work correctly you don't want the `src/huntsman/__init__.py` but I'll check (github doesn't let me comment on a file rename).

setup.cfg
danjampro: Yeah, `pip install -e .` seems to work file.

setup.py
danjampro: There was an import error for `huntsman.pocs` in `setup.py`. I've changed it so that the version is loaded directly based on [this](https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package) stack overflow question.

PR #191, opened 2020-08-17
Fix Travis tests

danjampro:
The build error is:
```
error: certifi 2019.3.9 is installed but certifi>=2020.06.20 is required by {'matplotlib'}
```
Solved by upgrading `certifi` in `.travis.yaml`. 

Also incremented python version to `3.8` to resolve conflict with `panoptes-utils`.


wtgee:
Let's hold off on this until I finish ~#168~ #195

PR #192, opened 2020-08-18
Fix dome shutter KeyError

danjampro:
Fixes a problem where the dome code would crash during call to `keep_dome_open` if `status` was an empty dictionary.  This would result in the dome shutting by itself.


fergusL (approved):
LGTM

PR #194, opened 2020-08-19
Fix remove camera in prepare_cameras

danjampro:
Fixes a bug where `observatory.prepare_cameras` would crash if a camera was removed from `cameras` (an `OrderedDict`) while looping over its entries, causing `RuntimeError('OrderedDict mutated during iteration')`. The fix is to move it outside of the loop over `cameras` items.

Also added a test to make sure it works.

### DIFF
--- a/src/huntsman/pocs/observatory.py
+++ b/src/huntsman/pocs/observatory.py
@@ -441,6 +441,7 @@ class HuntsmanObservatory(Observatory):
 
         # Wait for cameras to be ready
         n_cameras = len(self.cameras)
+        cameras_to_drop = []
         self.logger.debug('Waiting for cameras to be ready.')
         for i in range(1, max_attempts+1):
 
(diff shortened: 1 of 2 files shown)


src/huntsman/pocs/observatory.py
```diff
@@ -476,6 +476,12 @@ def prepare_cameras(self, sleep=60, max_attempts=5, require_all_cameras=False):
                                   f'waiting another {sleep} seconds before checking again.')
                 time.sleep(sleep)
 
+        # Remove cameras that didn't become ready in time
+        # This must be done outside of the main loop to avoid a RuntimeError
+        for cam_name in cameras_to_drop:
+            self.logger.debug(f'Removing {cam_name} from {self} for not being ready.')
```
JAAlvarado-Montes, line 482: Just a question: what does this print in 'self'? is that the list of cameras or the class?

JAAlvarado-Montes (approved):
Just raised a question out of curiosity, but all looks good to me. Approving.

src/huntsman/pocs/observatory.py
danjampro: Hi @JAAlvarado-Montes , thanks for looking at this! In this case the function is a method of the `HuntsmanObservatory` class, so `self` will refer to a `HuntsmanObservatory` instance. 

You might find this helpful:  https://stackoverflow.com/questions/1535327/how-to-print-instances-of-a-class-using-print

PR #203, opened 2020-09-01
Flat field timeout & config options

danjampro:
Quick fix to enable flat field code to continue in case of a timeout error. Cameras that timeout (configurable using the `flat_field_timeout` config parameter) are dropped from the remaining flat-field sequence.


conf_files/huntsman.yaml
```diff
@@ -51,12 +51,16 @@ mount:
...
+observatory:
+  flat_fields:
+    flat_field_order:
+      - one
+      - deux
+      - drei
+      - quattro
+    max_num_exposures: 5
+    max_attempts: 10
+    max_exptime: 120
```
wtgee: It's a little weird to have this under the `observatory` key since the entire config is related to the observatory.  Can we just leave it at `flat_fields`?

src/huntsman/pocs/observatory.py
```diff
@@ -775,13 +785,17 @@ def _take_flat_observation(self, exptimes, observation, fits_headers=None, dark=
...
-            time.sleep(1)
-
+        timeout = max(exptimes.values()).to_value(u.second) + flat_field_timeout
+        self.logger.debug(f"Waiting for flat-fields with timeout of {timeout}.")
+        if not wait_for_events(camera_events, timeout=timeout, sleep_delay=1):
+            self.logger.error("Timeout while waiting for flat fields.")
+
+        # Remove camera_events that timed out, removing them from the remaining flat-fielding
+        camera_events = {cam_name: event for cam_name, event in camera_events.items(
+                         ) if event.is_set()}
```
wtgee, line 795:
```suggestion
        camera_events = {
            cam_name: event 
            for cam_name, event 
            in camera_events.items() 
            if event.is_set()
        }
```

wtgee (approved):
LGTM.  I'd prefer we change the config key as mentioned.

src/huntsman/pocs/observatory.py
danjampro: Ewwww I don't like this multi-line style very much!

PR #206, opened 2020-09-08
Byobu startup script

fergusL:
Byobu startup script that automates the setup of a main control, shutter control, weather monitoring, camera servers, camera logs, dome control log and data management windows.

Note that some of the Pis don't seem to have ssh keys setup/shared with control computer and are prompting for passwords. Pairing of the Musca/TinyOS also prompts for a password. Might need a way to read in control computer password without having to store it within the script/repo.

Also note that currently I have not got the camera servers starting in a byobu session due to the ssh password problem (currently would require a human to enter the password before any commands can be executed within the ssh session). Also not sure how well nested byobu sessions will work.

So not fully automated but handles most of the setup. The script is also a bit slow.


wtgee:
@fergusL does this just replace `scripts/byobu_start.sh`?

Also note that the newer docker versions will potentially eliminate a lot of this. Although this looks clean and nice so some of it could be incorporated easily.

fergusL:
@wtgee yea `scripts/byobu_start.sh` is now obsolete. Doing the script in python allows for reading device info from config so camera server setup can be dynamic based on current config. All good if it just gets replaced by docker automation, still might be useful for day time testers this week.

wtgee:
Can you remove the shell script as part of this if obsolete?

This looks way better, thanks!  I'll try to give a  proper review soon.

src/huntsman/pocs/utils/__init__.py
```diff
@@ -1,7 +1,7 @@
 import os
 import netifaces
-from pocs.utils import listify
-from pocs.utils.config import load_config as config_loader
+from panoptes.utils import listify
+from panoptes.utils.config.helpers import load_config as config_loader
```
wtgee: Since this is in the main utils module when it loads this probably can't go through until #196 goes through, sorry. :(

You said this was working but did you open up other pocs items?

scripts/byobu_startup.py
```diff
@@ -0,0 +1,379 @@
...
+import time
+
+from huntsman.pocs.utils import load_config
+
+WINDOWS = ["main-control",
+           "shutter-and-weather",
+           "camera-servers",
+           "camera-logs",
+           "dome-control",
+           "data managment"]
```
wtgee: If this is consistent with the others...

```suggestion
           "data-managment"]
```

scripts/byobu_startup.py
```diff
@@ -0,0 +1,379 @@
...
+                "cameras = create_cameras_from_config",
+                "mount = create_mount_from_config(config)",
+                "mount.initialize()",
+                "scheduler = create_scheduler_from_config(config)",
+                "observatory = HuntsmanObservatory(cameras=cameras, "
+                "mount=mount, scheduler=scheduler, with_autoguider=True,"
+                " take_flats=True",
+                "pocs = POCS(observatory, simulator=['power','weather']",
+                "#pocs.initialize()",
+                "#pocs.run()"]
```
wtgee, line 35: This kind of gets lost in the commands.

```suggestion
                # Uncomment the following lines for automatic running
                "#pocs.initialize()",
                "#pocs.run()"]
```

wtgee (commented):
@fergusL this is way better than the current shell script in terms of readability and maintainability, thanks! I think it will definitely still be useful going forward. I'm giving my unsolicited review :grin: .

The big suggestion I would say is that you are doing a lot of repeating of `subprocess.call("byobu SOMETHING", shell=True)`.  Usually that much repetition of the same pattern is a sign (DRY - Don't Repeat Yourself) that you should pull it out into a function, sort of like you've already done with `send_command_to_pane` (and others), you just now need to go one level higher.   

For instance, if you needed to change to `shell=False` it's an easy search and replace, but if you decided you wanted to log each call to byobu it becomes more annoying. Factor out the actual `subprocess.call` and you have a lot more flexibility:

```py
def call_byobu(cmd, screen_cmd='byobu', shell=True):
    """Calls the given command within a byobu screen session.
    """
    run_cmd = f'{screen_cmd} {cmd}'
    subprocess.call(run_cmd, shell=shell)


...
# Somewhere else
call_byobu("new-window")
window_name = "Wow! What a cool window!"
call_byobu(f"rename-window '{window_name}'")
```

I didn't give a super thorough review of all of the actual commands that get run so I'm just leaving this review as a comment.

wtgee (approved):
I'm merging this without a thorough review but will be using it for live testing.  Expect a follow-up soon! ;)

Issue #205, opened 2020-09-03
Camera errors not handled by Huntsman "observing" state. 


danjampro:
If a timeout error (or maybe any other error) is encountered by the camera itself, the error does not get raised at the state machine level. This is potentially a fatal bug because it can cause the "observing" state to hang indefinitely.

wtgee:
Do you have anything from the logs that shows this happening?

danjampro:
@wtgee 

Camera logs:
```
D0903 10:09:15.884                 libasi.py:238] Set EXPOSURE to 300.0 s on camera 0
D0903 10:09:15.885                 libasi.py:450] Exposure started on camera 0
E0903 10:14:45.893                  error.py:016] Timeout: Timeout waiting for exposure on ZWO ASI Camera (361d420013090900) with Birger Focuser & ZWO Filter Wheel to complete
E0903 10:14:45.894                 camera.py:598] Error while waiting for exposure on ZWO ASI Camera (361d420013090900) with Birger Focuser & ZWO Filter Wheel: Timeout: Timeout waiting for exposure on ZWO ASI Camera (361d420013090900) with Birger Focuser & ZWO Filter Wheel to complete
```

And lots of 
```
Waiting for images ...
```
in the main logs.

PR #208, opened 2020-09-12
WIP: Improved camera error handling by adding new camera & focuser property to Pyro interface

AnthonyHorton:
Add new camera and focuser properties to the Pyro interface so that they can be used in huntsman-pocs.

Specifically:

1. Camera `.can_take_internal_darks` property (read only). This isn't actually anything to do with error handling but it's in POCS `develop` now so should be accessible for huntsman-pocs remote cameras too.
1. Camera `.exposure_error` property (read only). This will be `None` unless the most recent exposure taken by the camera raised an exception, in which case it will be equal to the string representation of that exception. By checking this after an exposure (or observation) has completed it is possible for calling code to detect cameras errors and respond accordingly, e.g. deleting failed cameras from the camera list.
1. Focuser `.autofocus_make_plots` property (read/write), can be used to check or set whether to produce the autofocus plots. This can still be overriden by setting the `make_plots` argument of the call to `autofocus()`.
1. Focuser `.autofocus_error` property (read only). This will be `None` unless the most recent autofocus taken by the focuser raise an exception, in which case it will be equal to the string representation of that exception. By checking this after an autofocus sequence has complete it is possible for calling code to detect camera errors and respond accordingly, e.g. deleting failed cameras from the camera list. 

**Note:** Apart from `.can_take_internal_darks` these new properties do not exist in POCS `develop`, only in the https://github.com/AnthonyHorton/POCS/tree/error-handling branch. See PR https://github.com/panoptes/POCS/pull/1003 for more details.

### DIFF
--- a/src/huntsman/pocs/camera/pyro.py
+++ b/src/huntsman/pocs/camera/pyro.py
@@ -122,6 +122,22 @@ class Camera(AbstractCamera):
         '''
         return self._proxy.get("is_ready")
 
+    @property
+    def can_take_internal_darks(self):
+        """ True if the camera can take internal dark exposures.
+        This will be true of cameras that have an internal mechanical shutter and can
+        be commanded to keep that shutter closed during the exposure. For cameras that
+        either lack a mechanical shutter or lack the option to keep it closed light must
+        be kept out of the camera during dark exposures by other means, e.g. an opaque
+        blank in a filterwheel, a lens cap, etc.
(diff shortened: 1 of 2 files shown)


AnthonyHorton:
These changes should at least partially fix #205.

danjampro:
@AnthonyHorton Is this still under development?

PR #209, opened 2020-09-12
Raspberry Pi cloud-init files.

### LINKED ISSUE #218: running checklist of things to do to setup an automated pi setup with cloud init
Supporting #209 , these are things that the team can help @wtgee with to setup this scheme:

- [ ] zshrc file 
- [ ] use shared `startup.sh` on the ssh mount to customise the startup of each pi (e.g. pull latest docker compose) https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot
- [ ] setup ssh keys
- [ ] auto mount control computer ssh share
- [ ] setup cron jobs (not sure what they are)
- [ ] other post-install commands
- [ ] point to new docker container?
- [ ] filling in all placeholders in current init file
- [ ] extras, e.g. set up cron jobs (downloading IERS for astroplan), auto-start camera services (?), etc.

wtgee:
These are the files are meant to be placed on a newly imaged raspberry Pi running ubuntu-server 64bit.

The instructions given by ubuntu should be followed: https://ubuntu.com/tutorials/how-to-install-ubuntu-on-your-raspberry-pi#1-overview

Section 3 of that guide talks about editing and moving the `network-config` file.  In our case we want to move all of these files.

So,

1. Install Ubuntu onto card via Imager on laptop.
2. When done, remove and reinsert SD card to laptop.
3.Change the `instance_id` in `meta-data` to be the camera name.
4. Change the `hostname` in `user-data` to be the camera name.
5. Add the NTP server at the top of the list in `user-data`.
4. Copy `user-data`, `meta-data`, `network-config`, and `usercfg.txt` to  `/system-boot` of the newly mounted SD card.
5. Unmount SD card and put in camera pi.
6. Boot pi.
7. Wash, rinse, repeat 1-7 for all cameras.
8. Profit. :pie: :money_mouth_face: 

This assumes the Pis have ethernet connection while being installed.  Wifi connection is not as simple as just editing the `network-config` if we want the other things in `user-data` to work.

In the future we could feasibly host an actual `cloud-init` instance (locally on the control computer), and then we would just have to edit one file, point it to the cloud, and it downloads the remaining files. It's more dynamic that way but not necessary for us.

Note that these are not complete yet.  We can set up all the shared ssh in here, install packages, etc. See [more examples here for ideas](https://cloudinit.readthedocs.io/en/latest/topics/examples.html).


resources/rpi/user-data
```diff
@@ -0,0 +1,74 @@
...
+hostname: huntsmanpi000
+
+# If you have set up ssh on github you can pull down your
+# key automatically so that you can log into the unit without
+# a password. Change 'CHANGEME' to your username below.
+ssh_import_id:
+  - gh:huntsmanscope
+
+chpasswd:
+  expire: true
```
AnthonyHorton, line 24: Is this enabling password expiry?

resources/rpi/user-data
```diff
@@ -0,0 +1,74 @@
...
+
+## Update apt database and upgrade packages on first boot
+package_update: true
+package_upgrade: true
+
+byobu: enable
+
+ntp:
+  enabled: true
+  servers:
```
AnthonyHorton: Might want to put the SSO ntp server in here, top of the list.

resources/rpi/user-data
```diff
@@ -0,0 +1,74 @@
...
+    - time1.google.com
+    - time2.google.com
+    - time3.google.com
+    - time4.google.com
+
+## Install additional packages on first boot
+packages:
```
AnthonyHorton, line 52: Do we want to include the non-free/binary SoC drivers that are mentioned in the (apparently really, really out of date) Ubuntu Raspberry Pi wiki? I know we're running headless so maybe it doesn't matter.

resources/rpi/usercfg.txt
```diff
@@ -0,0 +1,31 @@
+# Place "config.txt" changes (dtparam, dtoverlay, disable_overscan, etc.) in
+# this file. Please refer to the README file for a description of the various
+# configuration files on the boot partition.
+#
+# https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README
+# PANOPTES POCS Pi customisation for Raspberry Pi 4B+
+
+dtoverlay=disable-bt
+dtoverlay=i2c-gpio
```
AnthonyHorton, line 9: Only the domepi needs GPIO, but it probably doesn't hurt to have it enabled on the others too.

AnthonyHorton (approved):
Looks good to me, at least as a starting point.

resources/rpi/user-data
wtgee: It causes it to expire immediately so you are forced to change the password on first login.  I'll add a comment.

resources/rpi/user-data
wtgee: Actually, I did in the first commit and then got worried @lspitler wouldn't like that. :shushing_face:

resources/rpi/user-data
wtgee: I think we'll probably fill this out a lot more. There's also the option below to run scripts, so we could just have the install script that changes.

wtgee:
FYI, I think we will want to fix up the ssh first to use this.

resources/rpi/user-data
AnthonyHorton: Oh yeah, IPs in public repos. Not good.

wtgee:
FYI, this PR is editable so feel free to just make changes.  You can do it directly here in github if you don't want to checkout, etc.

resources/rpi/user-data
```diff
@@ -0,0 +1,76 @@
...
+# https://cloudinit.readthedocs.io/
+
+# Set the first part of the name to camera device.
+hostname: huntsmanpi000
+
+# If you have set up ssh on github you can pull down your
+# key automatically so that you can log into the unit without
+# a password. Change 'CHANGEME' to your username below.
+ssh_import_id:
+  - gh:huntsmanscope
```
wtgee, line 20: @lspitler do you own the `huntsmanscope` user?  We can do make some things slightly easier with ssh keys if you have a public ssh for that account.  If not we can create one.

danjampro (commented):
@wtgee Looks good - thanks for doing this! Do you think we should install the camera and filter wheel libraries on first boot?

wtgee:
> @wtgee Looks good - thanks for doing this! Do you think we should install the camera and filter wheel libraries on first boot?

We want to do as much as absolutely possible in these files so that we don't have to do any manual work on the Pi once it is booted and so that they are as consistent as they can be.

JAAlvarado-Montes:
Moved ntp info to the top of the file

lspitler:
On Pi8, I've setup further from the steps at top of PR using https://github.com/AstroHuntsman/huntsman-pocs/wiki/Setup-raspberry-Pi-as-a-camera-server#setting-up-a-docker-version. This is what I had to do:

```
sudo apt-get update
sudo apt-get -y install git python3-pip byobu vim grc
curl -fsSL get.docker.com -o get-docker.sh && sh get-docker.sh
apt-get install docker-compose
```

When ready to run docker:

```
docker pull huntsmanarray/camera:[github_huntsman_pocs_branch_name]
wget https://raw.githubusercontent.com/wtgee/huntsman-pocs/add-rpi-cloud-init-files/docker/camera/docker-compose.yml
```

Then just:

```
docker-compose run camera
```

Non-urgent, useful things:

- passwordless sudo

wtgee:
> Non-urgent, useful things:
> 
>     * passwordless sudo

This was not working?  It is set up in the `user-data` file already should already be done.

lspitler:
[setup and test cloud init](https://trello.com/c/wZKp2f9B/510-setup-and-test-cloud-init)

wtgee:
Time sync - chrony

lspitler:
@wtgee @danjampro https://trello.com/c/78SKh7oi/383-put-all-devices-onto-sso-ntp-server-time

lspitler:
Closing as https://github.com/AstroHuntsman/huntsman-pocs/pull/246 and https://github.com/panoptes/POCS/pull/1030 supersede this PR.

PR #215, opened 2020-09-17
Add feature for taking a different number of darks per exptime.

JAAlvarado-Montes:
Hi, 

This PR includes some modifications that allow `take_dark_fields` to take a different number of darks per exptime. This may come in handy when we want a variable number of darks, i.e. for small exptimes we would ideally want more darks, but for larger exptimes we would want to limit to a small number of darks.

Please review and let me know if everything's ok.

PS. This was tested on the control computer and it works perfectly.

Cheers,
Jaime.

### DIFF
--- a/src/huntsman/pocs/observatory.py
+++ b/src/huntsman/pocs/observatory.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import time
+import numpy as np
 from contextlib import suppress
 from functools import partial
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


wtgee:
I'm not entirely understanding the use case. Seems like that logic should be handled by a loop around this method rather than pulling it into the method itself.  Can you post an example of how you are using this?

JAAlvarado-Montes:
Well, that just means writing more and more code by separate which I think might be inconvenient. Currently, one only needs to do this and that's it:

`take_darks_from_config('FRB190608', 'pi0', exptimes=[1*u.s, 60*u.s, 0.0005*u.s,0.002*u.s], n_darks=[20, 10, 20, 20])`

So you can take a different number of darks per exptime. If you don't parse the arg `n_darks` then by default it will take 10 for each exptime.

lspitler (approved):

wtgee:
Um, well, I guess I won't review...

wtgee:
But, this needed better docstrings explaining the new features and you could include an example in the docstring otherwise the usage is odd.

You also don't need numpy for this operation, you can just do `[n_darks] * len(exptimes)`

And should ideally be using `listify` but it's not terribly important.

And I was going to suggest renaming `n_darks` to `num_darks`.

JAAlvarado-Montes:
OK @wtgee. Didn't know that `[n_darks] * len(exptimes)` would give me a list of the same len than exptimes, that's cool. I'll change that! Thanks!

PR #216, opened 2020-09-17
initial script for moving fits files from the POCS output directory to an archive directory

lspitler:
Aim is to determine when new fits files are moved to an "archive" space, where nifi processes and sends them to AAODC.

This allows us to delay this process significantly (i.e. if we're commissioning and want to view data on the control computer shortly after its taken) or not at all (i.e. we're in fully automated mode). 

A delay of a few seconds (at least) also has the benefit of ensuring nifi will never syncs both a compressed and  uncompressed version of a fits file, which is problematic because there are duplicate fits files in our AAODC database. Note, this might not be an issue in the future due to planned upgrades to POCS.

### DIFF
--- a/scripts/archive_fits_images.py
+++ b/scripts/archive_fits_images.py
@@ -21,10 +21,14 @@ from os import path
 @click.option('--test-mode',
               default=True,
               help='If true, move commands are printed to stdout, not run.')
+@click.option('--loop',
+              default=False,
+              help="If true, run a loop until ctrl+c")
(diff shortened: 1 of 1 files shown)


wtgee (approved):
Approving based off phone conversation with @lspitler and @danjampro

danjampro (approved):
Looks good to me. Maybe you can add a small description of why we need the script as a doc string at the top of the file?

PR #219, opened 2020-09-19
fixes, docstrings and upgrades to the archive fits script

lspitler:
Work in progress as we wrap up commissioning.


danjampro:
Closing as we no longer need/use this script.

PR #220, opened 2020-09-22
Byobu startup script updates

fergusL:
PR contains a few improvements that were suggested by @wtgee 
-added a `call_byobu()` function to replace the repeated `subprocess.call()` calls/make script easier to maintain
-fixed the problem that was leading to the creation of an extra unnamed byobu window (the `new_window()` function creates a new window and then renames it, so needed a function to just rename the current window and use that on the starting window before creating more)
-brought in some typo fixes/general modifications from the version on the control computer

### DIFF
--- a/scripts/byobu_startup.py
+++ b/scripts/byobu_startup.py
@@ -175,7 +175,7 @@ def create_12_pane_window(window_name, session_name="1-Huntsman-Control"):
     call_byobu(f"select-layout tiled")
 
 
-def setup_session(session_name="1-Huntsman-Control", windows=['']):
+def setup_session(session_name="1-Huntsman-Control", windows=None):
     """Create a new byobu session and populate desired windows.
 
     Parameters
@@ -186,9 +186,12 @@ def setup_session(session_name="1-Huntsman-Control", windows=['']):
         List of byobu window names.
 
     """
-    # subprocess.call(f"byobu new-session -d -s '{session_name}'", shell=True)
-    subprocess.call(f"byobu new-session -d -s'{session_name}'", shell=True)
-    for window in windows:
+    windows = windows or list()
+    call_byobu(f"new-session -d -s '{session_name}'")
+    # first we will rename the existing window
+    rename_window(windows[0])
+    # now iterate through list of desired window names and create them
+    for window in windows[1:]:
         new_window(window)
     select_window(windows[0])
     return
--- a/scripts/byobu_startup.py
+++ b/scripts/byobu_startup.py
@@ -64,7 +64,7 @@ def call_byobu(cmd, screen_cmd='byobu', shell=True, executable='/bin/zsh'):
 
     """
     run_cmd = f'{screen_cmd} {cmd}'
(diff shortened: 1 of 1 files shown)


fergusL:
@Physarah @JAAlvarado-Montes if there is anything that you think should be added to the byobu windows could you let me know so I can add/update the current setup?

scripts/byobu_startup.py
```diff
@@ -38,13 +38,35 @@
...
+    shell : bool
+        This argument sets whether to run the subprocess call cmd through a
+        shell, as opposed treating the input command as the name of an
+        executable to be run.
+    executable : sh
+        The path to the shell executable you wish to use.
+
+    """
+    run_cmd = f'{screen_cmd} {cmd}'
+    subprocess.call(run_cmd, shell=shell, executable='/bin/bash')
```
wtgee:
```suggestion
    subprocess.call(run_cmd, shell=shell, executable=executable)
```

scripts/byobu_startup.py
```diff
@@ -172,37 +209,37 @@ def setup_main_control_window(cmd_prefix='#'):
...
     send_command_to_pane(
-        cmd_prefix + '$HUNTSMAN_POCS/scripts/pyro_name_server.py', 0)
+        cmd_prefix + '${HUNTSMAN_POCS}/scripts/pyro_name_server.py', 0)
 
     send_command_to_pane(
-        cmd_prefix + 'python $HUNTSMAN_POCS/scripts/start_config_server.py', 1)
+        cmd_prefix + 'python ${HUNTSMAN_POCS}/scripts/start_config_server.py', 1)
 
-    send_command_to_pane(f'ipython', 2)
+    send_command_to_pane(f'jupyter-lab', 2)
```
wtgee: I'm not sure if we've decided this is what we want actually. Might be better to stick with `ipython`.

scripts/byobu_startup.py
```diff
@@ -172,37 +209,37 @@ def setup_main_control_window(cmd_prefix='#'):
...
-        time.sleep(0.01)
+    #for cmd in POCS_STARTUP:
+    #    send_command_to_pane(cmd_prefix + cmd, 2)
+    #    # issues occur without small pause between commands, unsure why
+    #    time.sleep(0.01)
 
-    send_command_to_pane(
-        cmd_prefix + 'grc tail -F -n 1000 $PANDIR/logs/ipython-all.log', 3)
+    #send_command_to_pane(
+    #    cmd_prefix + 'grc tail -F -n 1000 $PANDIR/logs/ipython-all.log', 3)
```
wtgee: If this is going to stay commented let's remove. Otherwise shouldn't the "dry run" flag make it so that it skips this?

wtgee (approved):
LGTM, thanks! Minor suggestions added.

scripts/byobu_startup.py
fergusL: I think this was something that was changed on the control computer and I just copied it over. I'm not sure why someone chose to comment it out

PR #224, opened 2020-10-13
Flat field config

danjampro:
- Merges `flat_field` and `flat_fields` config items
- Improves formatting in `huntsman.yaml`


bazkiaei (approved):
Looks good to me!

PR #225, opened 2020-10-13
Scheduler create function

wtgee:
Adds the scheduler creation helper from the DWF run. We will need this until #196 is complete.

Differs from the POCS version in how the config is looked up.

We used this during the DWF run, so I assume it's working? :thinking:


wtgee:
@danjampro looking at this I don’t think it’s needed, at least not here. The only difference is that it is manually loading the config files rather than using the config_server. That need should go away with the new merges.

Closing.

PR #228, opened 2020-10-21
Get tests passing

danjampro:
The goal of this PR is to get the config server and pyro services working out of the box (docker containers), with an additional aim to get tests running with a testing camera service. This is a priority task because there are several outstanding `huntsman-pocs` PRs that need to be tested before they are merged. Changes to the camera code can come later.

- A lot of test configuration moved to `docker/testing/docker-compose.yaml`.
- Local tests are now run using `scripts/testing/run_local_tests.sh`.
- pyro camera blocking is (temporarily) done on the service side until camera code gets revamped.
- Autofocus tests (and some others) are skipped until camera code gets revamped.
- Tests are done with only one camera for simplicity
- Camera service is not yet implemented as a `pytest` fixture, but this is on the to-do-list.
- Testing DSLR currently used because of bug in `POCS`.
- Small bug fixes related to dockerfile.
- Bug fixes inside `utils.pyro.service`.
- Lots of small changes to make things work...


docker/Dockerfile
```diff
@@ -17,10 +17,10 @@ ARG pip_extras="[testing]"
...
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 
-ENV PANDIR $pan_dir
-ENV PANLOG "$pan_dir/logs"
-ENV PANUSER $panuser
-ENV POCS $pocs_dir
+ENV PANDIR=$pan_dir
+ENV PANLOG="$pan_dir/logs"
+ENV PANUSER=$panuser
+ENV POCS=$pocs_dir
```
wtgee: Was this not working? Just curious.

.github/workflows/pythontest.yaml
```diff
@@ -22,30 +22,31 @@ jobs:
...
           name: codecov-upload
-          file: logs/coverage.xml
+          file: ${{ env.PANLOG }}/huntsman-coverage/coverage.xml
           fail_ci_if_error: true
-      - name: Create log file artifact
-        uses: actions/upload-artifact@v1
-        if: always()
-        with:
-          name: log-files
-          path: logs/huntsman-testing.log
```
wtgee: This would probably still be useful to have.

docker/docker-compose.yaml
```diff
@@ -49,21 +49,20 @@ services:
...
       - huntsman_dir:/var/huntsman
-    command: [ "/var/panoptes/POCS/bin/wait-for-it.sh localhost:6564 -- pyro5-httpgateway -e '.*'" ]
+    command: [ "/var/panoptes/POCS/scripts/wait-for-it.sh localhost:6564 -- pyro5-httpgateway -e '.*'" ]
 volumes:
   pandir:
     driver: local
     driver_opts:
       type: none
-      device: /var/panoptes
+      device: ${PANDIR}
```
wtgee, line 61: Using env vars here consistently gave me problems. This is working as expected?

scripts/setup-local-environment.sh
```diff
@@ -69,5 +69,5 @@ Done building the local images.
 
 To run the tests enter:
 
-scripts/testing/test-software.sh
+scripts/testing/test-docker-software.sh
```
wtgee: I think this is correct with the changes you've made:

```suggestion
scripts/testing/run_local_tests.sh
```

scripts/testing/run_docker_tests.sh
```diff
@@ -1,38 +1,44 @@
...
 PYRO_NS_HOST="$(panoptes-config-server get pyro.nameserver.host)"
 PYRO_NS_PORT="$(panoptes-config-server get pyro.nameserver.port)"
-
 echo "Starting the Pyro nameserver on ${PYRO_NS_HOST}:${PYRO_NS_PORT}"
 huntsman-pyro --verbose --host "${PYRO_NS_HOST}" --port "${PYRO_NS_PORT}" nameserver --auto-clean 300 &
 
+# Start a pyro camera service
+# TODO: Move inside conftest
+echo "Creating testing Pyro CameraService"
+huntsman-pyro --verbose service --service-name dslr.00 --service-class huntsman.pocs.camera.pyro.service.CameraService &
```
wtgee, line 32: This was hanging for me despite the `&`, which was part of the problem.  Is that still an issue?  Or was this the name change to `dslr.00` that fixed it?

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -104,6 +112,14 @@ def cooling_power(self):
...
         return self._proxy.get("is_exposing")
 
+    @is_exposing.setter
+    def is_exposing(self, is_exposing):
+        """Set or clear the remote exposure event."""
+        if is_exposing:
+            self._exposure_event.set()
+        else:
+            self._exposure_event.clear()
+
```
wtgee, line 122: Is this necessary to make things work or was this just part of trying to figure things out?

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -246,16 +243,14 @@ def autofocus(self, blocking=False, *args, **kwargs):
         self.logger.debug(f'Starting autofocus on {self}.')
 
         # Remote method call to start the exposure
-        self._proxy.autofocus(*args, **kwargs)
-
-        # Proxy for remote _autofocus_event
-        self._autofocus_event = RemoteEvent(self._proxy, event_type="focuser")
+        # self._proxy.autofocus(*args, **kwargs)
```
wtgee:
```suggestion
```

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -246,16 +243,14 @@ def autofocus(self, blocking=False, *args, **kwargs):
...
+        self._proxy.autofocus(*args, **kwargs, blocking=blocking)
```
wtgee:
```suggestion
        self._proxy.autofocus(blocking=blocking, *args, **kwargs)
```

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -269,33 +264,45 @@ def _readout(self, **kwargs):
...
+    def _run_timeout(self, event_name, event_type, blocking, max_wait):
         if blocking:
-            success = relevant_event.wait(timeout=max_wait)
-            if not success:
-                self._timeout_response(timeout_type, relevant_event)
+            event = getattr(self, f"_{event_name}_event")
+            timer = CountdownTimer(duration=max_wait)
+            while not timer.expired():
+                if not event.is_set():
+                    return
```
wtgee, line 272: This logic is still odd to me and I wonder if we need comments to explain?

src/huntsman/pocs/camera/pyro/service.py
```diff
@@ -115,3 +130,32 @@ def event_is_set(self, event_type):
...
+        """
+        device_configs = self.config.get("cameras")["devices"]
+
+        # Try and use the device name to select the device config
+        if device_name is not None:
+            for config in device_configs:
+                if config["name"] == device_name:
+                    self.logger.debug(f"Found camera config by name for {device_name}.")
+                    return config
+            self.logger.debug(f"Unable to find config entry for {device_name}.")
```
wtgee, line 152: You should be able to just query the device directly:

```suggestion
        device_config = self.config.get(f"cameras.devices.{device_name}")

        if device_config is not None:
            self.logger.debug(f"Found camera config by name for {device_name}.")
            return device_config
            
        self.logger.debug(f"Unable to find config entry for {device_name}.")
```

src/huntsman/pocs/camera/pyro/service.py
```diff
@@ -115,3 +130,32 @@ def event_is_set(self, event_type):
...
+
+        # If no match for device name, attempt to use IP address
+        ip_address = get_own_ip()
+        self.logger.debug(f"Querying for camera config with identifier: {ip_address}.")
+        for config in device_configs:
+            if config["name"] == ip_address:
+                return config
```
wtgee, line 159: Same as above:

```suggestion
        ip_address = get_own_ip()
        self.logger.debug(f"Querying for camera config with identifier: {ip_address}.")
        device_config = self.config.get(f"cameras.devices.{ip_address}")
        if device_config is not None:
            self.logger.debug(f"Found camera config by ip address for {device_name}.")
```

src/huntsman/pocs/camera/pyro/service.py
```diff
@@ -115,3 +130,32 @@ def event_is_set(self, event_type):
...
+
+        raise RuntimeError(f"Unable to find camera config entry for {ip_address}.")
```
wtgee, line 161:
```suggestion
        raise RuntimeError(f"Unable to find camera config entry for {device_name}.")
```

src/huntsman/pocs/camera/utils.py
```diff
@@ -128,18 +148,8 @@ def list_distributed_cameras(ns_host=None):
...
-            if n_cameras > 0:
-                logger.debug(f"Found {n_cameras} distributed cameras on name server")
-            else:
-                logger.warning(f"Found name server but no distributed cameras")
-    except error.PyroNameServerNotFound as err:
-        logger.warning(f"Couldn't connect to Pyro name server: {err!r}")
-        camera_uris = OrderedDict()
-
+    with get_running_nameserver() as name_server:
+        camera_uris = name_server.yplookup(meta_all=metadata)
```
wtgee, line 152: :question: 
```suggestion
        camera_uris = name_server.iplookup(meta_all=metadata)
```

src/huntsman/pocs/tests/test_huntsman.py
```diff
@@ -85,356 +85,3 @@ def pocs(config_with_simulated_stuff, observatory):
     yield pocs
 
     pocs.power_down()
-
```
wtgee, line 88: Are we not wanting any of this anymore? If we just want to skip them let's mark the whole file to skip rather than delete it.

src/huntsman/pocs/utils/config.py
```diff
@@ -17,3 +17,64 @@ def get_config(key, host=None, port=None, *args, **kwargs):
         *args,
         **kwargs
     )
+
+
+def get_own_ip(verbose=False, logger=None):
```
wtgee, line 22: Pyro5 has tools for finding the [ip address](https://pyro5.readthedocs.io/en/latest/servercode.html#network-adapter-binding-and-localhost). Is this function doing anything unique for us? I'd rather not maintain code that looks up an IP address if we don't have to.

src/huntsman/pocs/utils/pyro/cli.py
```diff
@@ -51,16 +51,17 @@ def nameserver(context, auto_clean=0):
...
                    'If the default `None`, then the device hostname will be used.')
 @click.option('--service-class', required=True, default=None,
               help='The class to register with Pyro. '
                    'This should be the fully qualified namespace for the class, '
                    'e.g. huntsman.pocs.camera.pyro.CameraService.')
+
+
 @click.pass_context
-def service(context, service_name, service_class=None):
+def service(context, service_name, service_class=None, config_identifier=None):
```
wtgee: You've added `config_identifier` but I don't see it being used?

src/huntsman/pocs/utils/pyro/nameserver.py
```diff
@@ -55,14 +55,15 @@ def pyro_nameserver(host=None,
             nameserver was started via `autostart=True`, the function will block until terminated,
             but still return the completed process.
     """
-    logger.info(f'Pyro nameserver start request: {host=} {port=} {auto_clean=} {auto_start=}')
+    logger.info(f'Pyro nameserver start request: host={host} port={port} auto_clean={auto_clean} \
+                auto_start={auto_start}.')
```
wtgee: Is this breaking up properly?

src/huntsman/pocs/observatory.py
```diff
@@ -207,7 +205,7 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None,
             safety_func = partial(self.is_dark, horizon='flat')
 
         # Load the flat fielding config
-        flat_field_config = self.get_config('flat_fields')
+        flat_field_config = self.get_config('flat_fields', default=dict())
```
wtgee, line 208: I think we have some inconsistency between `flat_fields` and `flat_field`. They are changed here to the `flat_fields` but the config still has `flat_field`.

wtgee (approved):
Thanks @danjampro! I did a pretty quick review of this as I know there's a few things flying around in here.

src/huntsman/pocs/camera/pyro/client.py
AnthonyHorton: I suspect this is rendered (almost) unnecessary by my changes in https://github.com/panoptes/POCS/pull/1026

Looks like a workaround for the `Camera` base class (and hence the Pyro client) trying to assign to `is_exposing` when as far as the Pyro service (and SDK cameras) are concerned `is_exposing` is read only.

src/huntsman/pocs/utils/config.py
AnthonyHorton: If Pyro5 has an equivalent for this I'm happy for this to be removed. It's not only less code to maintain, it also removed an external dependency (netifaces).  Back when I wrote this function there weren't any alternative solutions in Pyro or our other existing code/libraries.

src/huntsman/pocs/observatory.py
danjampro: Yep. I've changed it for the testing config but still need to update the huntsman config. Not sure whether to leave that as another PR though as I'm sure there will be other issues.

src/huntsman/pocs/utils/pyro/cli.py
danjampro: This is something I was experimenting with but is no longer needed. Good spot!

src/huntsman/pocs/utils/pyro/nameserver.py
danjampro: Probably not.

src/huntsman/pocs/utils/config.py
danjampro: Yep, sounds good. Let's worry about this in a separate PR though.

src/huntsman/pocs/tests/test_huntsman.py
danjampro: @wtgee The reason I deleted these is a) they are outdated and designed to test with the old config system and b) they seem to be mainly testing `pocs` functionality, rather than `huntsman-pocs`. Do we need this code hanging around?

src/huntsman/pocs/camera/pyro/client.py
danjampro: I think it was necessary to get the tests passing, but let's worry about this when we revisit the camera class.

scripts/testing/run_docker_tests.sh
danjampro: Doesn't seem to be an issue - I haven't seen it hang.

docker/docker-compose.yaml
danjampro: Seems to be working. Although these tests don't really cover this compose file. I'm hoping to get around to testing this on the real hardware ASAP.

.github/workflows/pythontest.yaml
danjampro: Sure, not sure why this got deleted. Where to these artefacts end up?

src/huntsman/pocs/camera/utils.py
danjampro: Weirdly enough, no. This stands for "yellow pages" lookup: https://pyro5.readthedocs.io/en/latest/nameserver.html

.github/workflows/pythontest.yaml
wtgee: They are in the Actions

Allow for arbitrary FITS headers (again) by wtgee · Pull Request #1016 · panoptesPOCS

src/huntsman/pocs/tests/test_huntsman.py
wtgee: I don’t think so and would prefer it was removed if not needed. I just wanted to check.

src/huntsman/pocs/utils/config.py
wtgee: Created #230

src/huntsman/pocs/camera/pyro/service.py
danjampro: @wtgee  The problem with this is that the device name is not the key, but rather a property of the device. This is by design in order to be consistent with `pocs`, which expects a list of devices rather than a dict.

src/huntsman/pocs/camera/pyro/client.py
AnthonyHorton: Yes, it would have been necessary, but what I'm suggesting is that now those other changes have been merged into POCS it's probably no longer necessary (and will probably actually break stuff).

PR #231, opened 2020-10-28
Fix docker ci

danjampro:
- Fixes broken docker CI by upgrading buildx action version (tested via GHA).
- Updates badges in readme.
- Renames actions for nicer badges.


wtgee (approved):
:+1: 

Do we know what changed between the v1 and v3 of the buildx helper that makes this work?

wtgee:
Only other comment is that we have an inconsistency with `yml` vs `yaml`. Not a huge deal but should probably make them consistent.

danjampro:
> 👍
> 
> Do we know what changed between the v1 and v3 of the buildx helper that makes this work?

To be honest, no. It was just a hunch I had given that it would successfully build locally with buildx.

danjampro:
> Only other comment is that we have an inconsistency with `yml` vs `yaml`. Not a huge deal but should probably make them consistent.

Good point. I've read that `.yml` is preferable over `.yaml`, but I'm not sure whether that would interfere with the way `pocs` is setup.

PR #233, opened 2020-11-03
Fix pyro fw

danjampro:
- Pyro FW code was broken and was not covered by units tests.
- Fixes pyro FW's remote event.
- Adds FW test but skips another that doesn't currently work.
- Camera / FW proxies are now properties that return new pyro proxies. This is to avoid thread ownership problems but may not be the most efficient solution.
- Fixes bug in `run_local_tests` script where volume was mounted incorrectly.


src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -46,6 +41,9 @@ def __init__(self, uri, name='Pyro Camera', model='pyro', port=None, *args, **kw
         self.connect()
 
     # Properties
+    @property
+    def _proxy(self):
+        return Proxy(self._uri)
```
wtgee, line 46: Since this is in fact a pyro class, I could make an argument for this being non-private (i.e. `proxy` vs `_proxy`).

src/huntsman/pocs/observatory.py
```diff
@@ -711,7 +711,6 @@ def _take_flat_observation(self, exptimes, observation, fits_headers=None, dark=
         self.logger.debug(f'Slewing to flat-field coords: {observation.field}.')
         self.mount.set_target_coordinates(observation.field)
         self.mount.slew_to_target()
-        self.status()  # Seems to help with reading coords
```
wtgee, line 714: Does this need to be removed?

tests/testing.yaml
```diff
@@ -85,6 +96,7 @@ flat_fields:
   max_num_exposures: 1
   max_attempts: 1
   max_exptime: 10
+  flat_field_timeout: 30
```
wtgee: This is under the `flat_fields` key so can it just be `timeout`?

wtgee (approved):
:+1:

src/huntsman/pocs/camera/pyro/client.py
danjampro: You have a point. Let's leave this for a later discussion.

src/huntsman/pocs/observatory.py
danjampro: It was raising errors because `self.status` returns a `dict` and is not callable. It shouldn't be necessary anyway. I've seen codes work without it, and tests pass, so let's get rid of it!

wtgee: I have tried to make all `status` a property rather than a method so if we can just switch to that it would be better.

wtgee: But if that involves doing something to make it work then fine to delete here and handle in separate PR.

PR #240, opened 2020-11-10
Fix pyro focusing

danjampro:
- Fixes the pyro code to get remote autofocusing behaving properly
- Cleans up the pyro code to remove unused functions
- Reinstates autofocus tests

### DIFF
--- a/src/huntsman/pocs/camera/pyro/client.py
+++ b/src/huntsman/pocs/camera/pyro/client.py
@@ -244,7 +244,7 @@ class Camera(AbstractCamera):
 
         if timeout is None:
             coarse_str = "coarse" if coarse else "fine"
-            timeout = self.config.get(f"focusing.{coarse_str}.timeout", default=600)
+            timeout = self.get_config(f"focusing.{coarse_str}.timeout", default=600)
 
         # Remote method call to start the exposure
         self.logger.debug(f'Starting autofocus on {self} with timeout: {timeout}.')
--- a/src/huntsman/pocs/camera/pyro/client.py
+++ b/src/huntsman/pocs/camera/pyro/client.py
@@ -229,7 +229,7 @@ class Camera(AbstractCamera):
                 False.
             blocking (bool, optional): Whether to block (on the client-side) until autofocus
                 complete, default False.
-            timeout (float, optional): The client-side autofocus timeout. Default 5 min.
+            timeout (float, optional): The client-side autofocus timeout. The default value of `None` will lookup the `focusing.<focus_type>.timeout` value in the config server. If not provided, a default fallback of 600 seconds is used.
 
         Returns:
             threading.Event: Event that will be set when autofocusing is complete
--- a/src/huntsman/pocs/camera/pyro/client.py
+++ b/src/huntsman/pocs/camera/pyro/client.py
@@ -266,7 +266,7 @@ class Camera(AbstractCamera):
             if not proxy.is_reading_out:
                 return
             timer.sleep(sleep_time)
-        raise error.PaneError(f"Timeout of {timeout} reached while waiting for readout to finish"
+        raise error.PanError(f"Timeout of {timeout} reached while waiting for readout to finish"
                               f" on camera client {self}.")
 
     def _start_exposure(self, **kwargs):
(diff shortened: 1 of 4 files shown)


src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -238,20 +242,32 @@ def autofocus(self, blocking=False, *args, **kwargs):
...
+        Args:
+            timeout (float): The readout timeout in seconds.
+        """
+        proxy = self._proxy
+        timer = CountdownTimer(timeout)
+        while not timer.expired():
+            if not proxy.is_reading_out:
+                return
+            timer.sleep(sleep_time)
+        raise error.PaneError(f"Timeout of {timeout} reached while waiting for readout to finish"
```
wtgee:
```suggestion
        raise error.PanError(f"Timeout of {timeout} reached while waiting for readout to finish"
```

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -261,41 +277,6 @@ def _readout(self, **kwargs):
...
-                    return
-            self._timeout_response(event_name, event_type, max_wait, blocking)
-
-        else:
-            # If the remote operation fails after starting in such a way that the event doesn't
-            # get set then calling code could wait forever. Have a local timeout thread
-            # to be safe.
-            timeout_thread = Timer(interval=max_wait, function=self._timeout_response,
-                                   args=(event_name, event_type, max_wait, blocking))
-            timeout_thread.start()
```
wtgee, line 279: Is this not needed any more?

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -225,7 +227,9 @@ def autofocus(self, blocking=False, *args, **kwargs):
                 a fine focus. Default False.
             make_plots (bool, optional: Whether to write focus plots to images folder, default
                 False.
-            blocking (bool, optional): Whether to block until autofocus complete, default False.
+            blocking (bool, optional): Whether to block (on the client-side) until autofocus
+                complete, default False.
+            timeout (float, optional): The client-side autofocus timeout. Default 5 min.
```
wtgee: We should come up with a standard way of documenting these "if `None` then check config else use fallback of 600 seconds" kind of statements.  The default here can change without any code changes, which is a bit misleading.  I'm guilty of this in many other places.

```suggestion
            timeout (float, optional): The client-side autofocus timeout. The default value of `None` will lookup the `focusing.<focus_type>.timeout` value in the config server. If not provided, a default fallback of 600 seconds is used.
```

src/huntsman/pocs/camera/pyro/service.py
```diff
@@ -76,17 +79,20 @@ def get_uid(self):
...
         return self._camera.uid
 
-    def take_exposure(self, *args, blocking=False, **kwargs):
-        # Start the exposure non-blocking so that camera server can still respond to
-        # status requests.
-        # self._exposure_event = self._camera.take_exposure(blocking=False, *args, **kwargs)
-        self._exposure_event = self._camera.take_exposure(*args, **kwargs, blocking=blocking)
+    def take_exposure(self, *args, **kwargs):
+        """
+        """
```
wtgee:
```suggestion
        """Proxy call to the camera client.
        
        This method will strip any `blocking` parameter that is passed so Pyro can handle the blocking appropriately.
        """
```

wtgee (approved):
LGTM. One question about the removal of the extra timeout checks.

src/huntsman/pocs/camera/pyro/client.py
danjampro: @wtgee  I'm not sure why it would be. The `_focus_event` is a `RemoteEvent` object, so its timeout is  implicitly handled on by the camera service. 

Otherwise, the class should probably try and mimic the behaviour of `AbstractCamera` as much as possible, which doesn't spawn a new thread to monitor its `readout_thread` in `take_exposure`.

I guess the only reason we would keep something like this is to raise an error/warning on the client side if there is a timeout on the camera service. But this would be raised inside a different thread anyway.

Thoughts?

PR #246, opened 2020-12-03
camera-pi cloud init

danjampro:
Cloud-init files for a huntsman camera pi. Everything required to run the docker camera service.

- Docker camera service is **automatically run on boot** inside a byobu window
- The most-recent docker images and docker-compose files are always used
- **We still need to specify `PANOPTES_CONFIG_HOST` and setup ssh keys manually**

- There are currently still some hardcoded references to `danjampro/huntsman-pocs` that will need to be removed before merging. I'm keeping them at the moment to continue testing.


docker/camera/docker-compose.yaml
```diff
@@ -1,20 +1,16 @@
...
-      PANOPTES_CONFIG_FILE:
+      PANDIR: /var/huntsman
       PANOPTES_CONFIG_HOST:
       PANOPTES_CONFIG_PORT:
     command: [ "huntsman-pyro --verbose service --service-class huntsman.pocs.camera.pyro.service.CameraService" ]
     volumes:
-      - "~/.ssh:/home/huntsman/.ssh"
-      - "/var/huntsman/images:/var/huntsman/images"
+      - "${PANDIR}/images:/var/huntsman/images"
+      - "${PANDIR}/logs:/var/huntsman/logs"
```
wtgee, line 16: Just a note that this is a bit confusing because the interpolation is actually happening on the host itself, not inside the running container.  So `$PANDIR` must exist on the host even though it is explicitly set on line 10 inside the container. So `$${PANDIR}/logs` here would make it use the env var inside the container, which is odd usage.

resources/rpi/user-data
```diff
@@ -0,0 +1,95 @@
...
+      user_allow_other
+    path: /etc/fuse.conf
+    append: true
+
+## Get and run the install script upon first boot.
+runcmd:
+  # Setup hardware watchdog
+  - echo 'watchdog-device = /dev/watchdog' >> /etc/watchdog.conf
+  - echo 'watchdog-timeout = 15' >> /etc/watchdog.conf
+  - echo 'max-load-1 = 24' >> /etc/watchdog.conf
```
wtgee: We might need to think about these a bit. When I tested it I was getting resets when I didn't want to.

resources/rpi/usercfg.txt
```diff
@@ -0,0 +1,29 @@
...
+dtparam=eth_led1=4
+
+# Disable activity LED
+dtoverlay=act-led
+dtparam=act_led_trigger=none
+dtparam=act_led_activelow=off
+
+# Disable power LED
+dtparam=pwr_led_trigger=none
+dtparam=pwr_led_activelow=off
```
wtgee: These probably need to be double-checked that they are working once installed.

scripts/camera/install-camera-pi.sh
```diff
@@ -0,0 +1,141 @@
...
+  # Set up autologin without password for huntsman user
+  # This is a bit of a hack but it appears to be the standard method of doing this
+  sed -i '/^ExecStart=$/d' /lib/systemd/system/getty@.service
+  sed -i "s/ExecStart=.*/ExecStart=\nExecStart=-\/sbin\/agetty -a huntsman --noclear %I \$TERM/g" /lib/systemd/system/getty@.service
+  sed -i "s/Type=idle/Type=simple/g" /lib/systemd/system/getty@.service
+
+  # We also need to disable password change on first login
+  # The easiest way of doing this is just to change the password here
+  # The password expiry status can be checked by doing $chage -l ${PAUSER}
+  usermod --password "${PANUSER}" "${PANUSER}"
```
wtgee, line 78: Is this still necessary? I saw that the expiration was removed from the `user-data` file.

scripts/camera/run-camera-service.sh
```diff
@@ -0,0 +1,49 @@
...
+
+clear
+echo "############### Huntsman Camera Service ###############"
+echo "Control computer hostname: ${REMOTE_HOST}"
+
+# Mount the SSHFS images directory
+echo "Mounting remote images directory ${REMOTE_IMAGES_DIR} to ${LOCAL_IMAGES_DIR}"
+mkdir -p ${LOCAL_IMAGES_DIR}
+sudo umount ${LOCAL_IMAGES_DIR} || true
+sshfs -o allow_other ${REMOTE_IMAGES_DIR} ${LOCAL_IMAGES_DIR}
```
wtgee: Does this stay connected if the byobu session dies?  Could we just alter fstab to have it auto mount on boot?

Also, there are various reconnect options that might be useful to specify. I will try to look up what I have used in the past.

scripts/camera/run-camera-service.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Get the docker-compose file
+DC_FILE_URL=https://raw.githubusercontent.com/${GITHUB_USER}/huntsman-pocs/${GITHUB_BRANCH}/docker/camera/docker-compose.yaml
+DC_FILE="${PANDIR}/docker-compose.yaml"
+if [ -f ${DC_FILE} ] ; then
+    echo "Removing existing docker-compose file: ${DC_FILE}"
+    rm ${DC_FILE}
+fi
+echo "Downloading latest docker-compose file from ${DC_FILE_URL} to ${DC_FILE}"
+wget ${DC_FILE_URL} -O ${DC_FILE}
```
wtgee: The fact that this all happens automatically is a bit odd to me, although I'm not sure why. Mostly the pull of the latest file from github upon each restart. But I haven't fully thought through what it should maybe be so this might make the most sense.

wtgee (approved):
Quick review, but LGTM in terms of changes. A few suggestions but I think most changes would come from testing on real hardware.

docker/camera/docker-compose.yaml
danjampro: Yes - this is intentional as we want those variables to be expanded on the host. Also with the cloud init setup we can guarantee that the PANDIR variable will be set. I explicitly set it to `/var/huntsman` inside the container because for some reason (I think it may be the `pocs` base image) it was getting set to `/var/panoptes`.

resources/rpi/user-data
danjampro: @wtgee should I remove them for now?

scripts/camera/install-camera-pi.sh
danjampro: I removed it from `user-data` but it seems that ubuntu's default behaviour is to require the immediate password reset - this is how I got around it.

scripts/camera/run-camera-service.sh
danjampro: Doing it on boot could be something to consider. And yes, I think we implemented those in the `huntsman-pocs` SSHFS implementation.

scripts/camera/run-camera-service.sh
danjampro: I guess the idea is just to guarantee we are using the latest docker image with the latest docker file.

PR #256, opened 2020-12-10
Stop restart of camera service

danjampro:
- Remove `unsetopt` which results in command not found
- Only attempt to start one camera service (based on presence of `huntsman` byobu window)
- Do not initially attach to `huntsman` window


JAAlvarado-Montes (approved):
LGTM!

PR #257, opened 2020-12-10
Parse camera config when creating distributed cameras

danjampro:
- Currently the camera config is not parsed to pyro camera clients, meaning we cannot control things like temperature tolerance
- Parse camera config when creating distributed cameras


JAAlvarado-Montes (approved):
LGTM!

PR #260, opened 2020-12-10
Mount docker logs

danjampro:
- Current control computer logs are only accessible from inside docker image
- This PR mounts the log directory so the logs are easily accessible from the control comp outside of docker
- Minor camera fix to allow a pyro camera instance to be "stringified"
- Pyro http server runs on port 8081 to avoid clash with lights


JAAlvarado-Montes (approved):
LGTM

PR #261, opened 2020-12-13
Added changes that were tested in the last DWF run for the taking_darks state

JAAlvarado-Montes:
Hi,

This PR includes all the changes that were made to the `taking_darks` state during the last DWF run in September, and they were used to collect a significant number of dark images. Main changes include:

- Move all filterwheels to `blank` position in `take_dark_fields` before exposing.
- Fixed bug in `taking_darks` state to grab the list of exposure times from `targets.yaml`.

All unit tests are passing successfully and all the changes have already been tested with real hardware. Everything is up-to-date with the latest `POCS` version.

Please let me know any comment or suggestion.

Cheers,
Jaime A.

### DIFF
--- a/src/huntsman/pocs/observatory.py
+++ b/src/huntsman/pocs/observatory.py
@@ -8,7 +8,7 @@ from astropy import stats
 from astropy import units as u
 from astropy.io import fits
 
-from panoptes.utils import error, altaz_to_radec, listify
+from panoptes.utils import error, altaz_to_radec, listify, get_quantity_value
 from panoptes.utils.library import load_module
 from panoptes.utils.time import current_time, flatten_time, wait_for_events
 
--- a/src/huntsman/pocs/observatory.py
+++ b/src/huntsman/pocs/observatory.py
@@ -321,8 +321,9 @@ class HuntsmanObservatory(Observatory):
             n_darks = listify(n_darks) * len(exptimes)
 
         if set_from_config:
-            exptimes = self.config["calibration"]["darks"]["exposure_time"]
-            n_darks = self.config["calibration"]["darks"]["n_darks"]
+            dark_config = self.get_config('calibration', default=dict())
+            exptimes = dark_config["darks"]["exposure_time"]
+            n_darks = dark_config["darks"]["n_darks"]
 
         self.logger.debug(f"Going to take {n_darks} dark-fields for each of these exposure times {exptimes}")
 
--- a/src/huntsman/pocs/observatory.py
+++ b/src/huntsman/pocs/observatory.py
@@ -280,16 +280,19 @@ class HuntsmanObservatory(Observatory):
 
         Args:
             exptimes (list, optional): List of exposure times for darks
-            sleep (float, optional): Time in seconds to sleep between dark sequences.
             camera_names (list, optional): List of cameras to use for darks
             n_darks (int or list, optional): if int, the same number of darks will be taken
                 for each exptime. If list, the len has to be the same than len(exptimes), where each
                 element is the number of darks we want for the corresponding exptime, e.g.:
-                take_dark_fields(exptimes=[1*u.s, 60*u.s, 15*u.s], n_darks=[30, 10, 20])
+                take_dark_images(exptimes=[1*u.s, 60*u.s, 15*u.s], n_darks=[30, 10, 20])
                 will take 30x1s, 10x60s, and 20x15s darks
-            set_from_config (bool, optional): flag to set exptimes and n_darks directly from config file.
+            set_from_config (bool, optional): flag to set exptimes and n_darks directly
+                from config file.
             imtype (str, optional): type of image
         """
+
+        if exptimes is None:
+            exptimes = list()
 
         if camera_names is None:
             cameras_list = self.cameras
@@ -304,15 +307,15 @@ class HuntsmanObservatory(Observatory):
 
         # Move all the camera filterwheels to the blank position.
         for camera in cameras_list.values():
-            if camera.filterwheel.current_filter != "blank":
-                filterwheel_event = camera.filterwheel.move_to("blank")
+            if camera.filterwheel.current_filter != 'blank':
+                filterwheel_event = camera.filterwheel.move_to_dark_position()
                 filterwheel_events[camera] = filterwheel_event
-        self.logger.debug('Waiting for all the filterwheels to be on blank filter...')
+        self.logger.debug('Waiting for all the filterwheels to move to the dark position')
         wait_for_events(list(filterwheel_events.values()))
 
         # List to check that the final number of darks is equal to the number
         # of cameras times the number of exptimes times n_darks.
-        darks_filenames = []
+        darks_filenames = list()
 
         exptimes = listify(exptimes)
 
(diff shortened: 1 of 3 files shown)


lspitler (approved):
LGTM from last reviews after DWF run (also good to start using this code ASAP as it fixes the directory structure). @JAAlvarado-Montes only two changes:

Can you rename every mention of "dark_field" to "dark_image" in the variables, function names, tests and docstrings? Field usually implies we're pointing at a "field" on the sky, which is not the case for darks.

What happens if someone gives `=[]` for exposure times and there is nothing in the `darks : exposure_time` configuration file?

src/huntsman/pocs/observatory.py
```diff
@@ -267,26 +267,28 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None, safety_func=Non
 
         self.logger.info('Finished flat-fielding.')
 
-    def take_dark_fields(self,
-                         exptimes,
+    def take_dark_images(self,
+                         exptimes=[],
```
danjampro: It is more common to have `exptimes=None` in the function arguments and then have a statement like
```
if exptimes is None:
    exptimes = list()
```
Somewhere near the start of the function.

src/huntsman/pocs/observatory.py
```diff
@@ -297,9 +299,17 @@ def take_dark_fields(self,
...
 
-        self.logger.debug(f"Taking {n_darks} dark exposures with exposure times: {exptimes}")
+        self.logger.debug(f'Moving all camera filterwheels to blank filter')
+
+        # Move all the camera filterwheels to the blank position.
+        for camera in cameras_list.values():
+            if camera.filterwheel.current_filter != "blank":
+                filterwheel_event = camera.filterwheel.move_to("blank")
+                filterwheel_events[camera] = filterwheel_event
+        self.logger.debug('Waiting for all the filterwheels to be on blank filter...')
```
danjampro: Change message to `Waiting for all the filter wheels to move to the dark position`

src/huntsman/pocs/observatory.py
```diff
@@ -297,9 +299,17 @@ def take_dark_fields(self,
...
 
-        image_dir = self.get_config('directories.images')
+        filterwheel_events = dict()
 
-        self.logger.debug(f"Taking {n_darks} dark exposures with exposure times: {exptimes}")
+        self.logger.debug(f'Moving all camera filterwheels to blank filter')
+
+        # Move all the camera filterwheels to the blank position.
+        for camera in cameras_list.values():
+            if camera.filterwheel.current_filter != "blank":
```
danjampro: We should be using the generic `dark_position` rather than explicitly using the `blank` filter name here. You can use this: `filterwheel.move_to_dark_position()`

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
         if not isinstance(n_darks, list):
             n_darks = listify(n_darks) * len(exptimes)
 
+        if set_from_config:
+            dark_config = self.get_config('calibration', default=dict())
```
danjampro: I think the dark config should be `self.get_config('calibs.dark', default=dict())`. Then we can set the flat fields up consistently so their config lives in `self.get_config('calibs.flats'). Then you wouldn't have to use the `dark` key in the lines below, which is a bit cleaner.

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
...
 
-            start_time = current_time()
+            for exptime, num_darks in zip(exptimes, n_darks):
+
+                start_time = current_time()
 
-            with suppress(AttributeError):
-                exptime = exptime.to_value(u.second)
+                with suppress(AttributeError):
+                    exptime = exptime.to_value(u.second)
```
danjampro: We should use `panoptes.utils.get_quantity_value` for this. Like
```
exptime = get_quantity_value(exptime, u.second)
```

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
...
+                # Create dark observation
+                dark_obs = self._create_dark_observation(exptime)
 
-            # Loop over exposure times for each camera.
-            for num in range(num_darks):
+                # Loop over exposure times for each camera.
+                for num in range(num_darks):
 
-                self.logger.debug(f'Darks sequence #{num} of exposure time {exptime}s')
+                    self.logger.debug(f'Darks sequence #{num} of exposure time {exptime}s')
```
danjampro: Change message to:
```
f"Taking {num} of {num_darks} dark frames with exposure time={exptime}.
```

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
...
-                                        dark_obs.seq_time)
+                        # Set header
+                        fits_headers = self.get_standard_headers(observation=dark_obs)
+                        # Common start time for cameras
+                        fits_headers['start_time'] = flatten_time(start_time)
 
-                    filename = os.path.join(
-                        path, f'{imtype}_{num:02d}.{camera.file_extension}')
+                        # Create filename
+                        path = os.path.join(dark_obs.directory,
```
danjampro: I think this can fit all on one line. The line limit for `panoptes` projects is 100 characters. Same for some of the other lines in this function.

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
...
-                # Block until done exposing on all cameras
-                while not all([info['event'].is_set() for info in camera_events.values()]):
-                    self.logger.debug('Waiting for dark-field images...')
-                    time.sleep(sleep)
-        self.logger.debug(darks_filenames)
-        return darks_filenames
+                        self.logger.debug(camera_events)
+
+                    # Block until done exposing on all cameras
+                    while not all([info['event'].is_set() for info in camera_events.values()]):
```
danjampro: We should be using `panoptes.utils.time.wait_for_events`  here.

danjampro (commented):
Hi @JAAlvarado-Montes. Looking good! I added a few suggestions, mostly cleanups and code-style changes. Let me know if you have any questions.

src/huntsman/pocs/observatory.py
```diff
@@ -310,66 +320,79 @@ def take_dark_fields(self,
...
             n_darks = listify(n_darks) * len(exptimes)
 
+        if set_from_config:
+            dark_config = self.get_config('calibration', default=dict())
+            exptimes = dark_config["darks"]["exposure_time"]
+            n_darks = dark_config["darks"]["n_darks"]
+
         # Loop over cameras.
-        for exptime, num_darks in zip(exptimes, n_darks):
+        if len(exptimes) > 0:
```
lspitler: @JAAlvarado-Montes how about have a if statement before this block instead so you don't have to indent this entire thing. I think you want this to error as well.

```suggestion
        if len(exptimes) == 0:
            raise PanError("No exposure times were provided. No dark images were taken.")
```

src/huntsman/pocs/observatory.py
```diff
@@ -267,26 +267,28 @@ def take_flat_fields(self, camera_names=None, alt=None, az=None, safety_func=Non
...
 
         Args:
-            exptimes (list): List of exposure times for darks
+            exptimes (list, optional): List of exposure times for darks
             sleep (float, optional): Time in seconds to sleep between dark sequences.
             camera_names (list, optional): List of cameras to use for darks
             n_darks (int or list, optional): if int, the same number of darks will be taken
                 for each exptime. If list, the len has to be the same than len(exptimes), where each
                 element is the number of darks we want for the corresponding exptime, e.g.:
                 take_dark_fields(exptimes=[1*u.s, 60*u.s, 15*u.s], n_darks=[30, 10, 20])
```
lspitler:
```suggestion
                take_dark_images(exptimes=[1*u.s, 60*u.s, 15*u.s], n_darks=[30, 10, 20])
```

src/huntsman/pocs/scheduler/dark_observation.py
```diff
@@ -7,7 +8,8 @@
 
 
 class DarkObservation(Observation):
-    """ A Dark-field observation
+
+    """ A Dark observation
 
     Dark observations will consist of multiple exposure. As the mount will be
     parked when using this class, the fields will be centred at the parked
```
lspitler:
```suggestion
    parked when using this class, the fits image header RA, Dec will be centred at the parked
```

lspitler (approved):
A few more @JAAlvarado-Montes

src/huntsman/pocs/observatory.py
lspitler: Or an error as I suggest below.

PR #262, opened 2020-12-14
Commissioning changes

danjampro:
This PR includes several changes that were required last week but not merged. They have been fully tested on Huntsman hardware. They require several outstanding POCS PRs to be merged in order to work properly on the hardware.

### DIFF
--- a/src/huntsman/pocs/camera/pyro/client.py
+++ b/src/huntsman/pocs/camera/pyro/client.py
@@ -310,7 +310,7 @@ class Camera(AbstractCamera):
                     fits.getdata(filename)
                     self.logger.debug(f"Finished waiting for file {filename}.")
                     return
-            timer.sleep(sleep_interval)
+            timer.sleep(sleep_interval)sada
         raise error.PanError(f"Timeout of {timeout} reached while waiting for file {filename} to"
                              f" exist on camera client {self}.")
 
--- a/src/huntsman/pocs/camera/pyro/client.py
+++ b/src/huntsman/pocs/camera/pyro/client.py
@@ -190,9 +190,8 @@ class Camera(AbstractCamera):
     def take_exposure(self, seconds=1.0 * u.second, filename=None, dark=False, blocking=False,
                       sleep_interval=0.5, max_write_time=10, *args, **kwargs):
         """Take an exposure for given number of seconds and saves to provided filename.
-
         Args:
-            seconds (u.second, optional): Length of exposure.
+            seconds (astropy.Quantity, optional): Length of exposure.
             filename (str, optional): Image is saved to this filename.
             dark (bool, optional): Exposure is a dark frame, default False. On cameras that support
                 taking dark frames internally (by not opening a mechanical shutter) this will be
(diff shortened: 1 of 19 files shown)


bazkiaei (approved):
Looks good to me!

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -183,21 +188,22 @@ def connect(self):
...
-            seconds (u.second, optional): Length of exposure.
+            seconds (astropy.Quantity, optional): Length of exposure.
             filename (str, optional): Image is saved to this filename.
             dark (bool, optional): Exposure is a dark frame, default False. On cameras that support
                 taking dark frames internally (by not opening a mechanical shutter) this will be
                 done, for other cameras the light must be blocked by some other means. In either
                 case setting dark to True will cause the `IMAGETYP` FITS header keyword to have
                 value 'Dark Frame' instead of 'Light Frame'. Set dark to None to disable the
                 `IMAGETYP` keyword entirely.
+            max_write_time (astropy.Quantity, optional): The maximum allowable delay between the
```
wtgee, line 202: `sleep_interval` needs to be documented.

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -274,20 +284,32 @@ def autofocus(self, blocking=False, timeout=None, coarse=False, *args, **kwargs)
...
+            timeout (float): The timeout in seconds.
         """
         proxy = self._proxy
         timer = CountdownTimer(timeout)
         while not timer.expired():
             if not proxy.is_reading_out:
-                return
+                try:
+                    # The best way of checking the file is written appears to be to get its data
+                    fits.getdata(filename)
```
wtgee, line 310: This does seem like an expensive way of checking for a valid file, especially in a loop. The data is also not accessed (it's memory mapped) by default, so not sure this is actually doing what you would want.

What is the error this is trying to solve? As in, where does the problem arise if the file is not valid?

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -274,20 +284,32 @@ def autofocus(self, blocking=False, timeout=None, coarse=False, *args, **kwargs)
...
+                    self.logger.debug(f"Finished waiting for file {filename}.")
+                    return
+                except (TypeError, FileNotFoundError):
```
wtgee: Can change this `try/except` to a `with suppress`

src/huntsman/pocs/states/huntsman/dithering.py
```diff
@@ -16,7 +16,6 @@ def on_enter(event_data):
 
         if pocs.observatory.mount.set_target_coordinates(current_observation.field):
             pocs.observatory.slew_to_target()
-            pocs.status()
```
wtgee, line 19: What's the reason for removing these?

wtgee (commented):
Can approve if needed. Curious about why the `status` calls were removed.

src/huntsman/pocs/states/huntsman/dithering.py
danjampro: @wtgee I'm not sure why they are there in the first place? And currently they are being called, but `pocs.status` returns a `dict`, so they are raising errors.

src/huntsman/pocs/camera/pyro/client.py
danjampro: @wtgee The problem is that there is a delay between the camera service saying it has written the image and it actually appearing as a complete FITS file on the control computer. So sometimes we try and read a partial FITS file. From my experience, the header can actually be read from an incomplete FITS file - calling `getadata` is (again from my experience) the best way of checking if the file is fully written because it uses the FITS `CHECKSUM`. If it is memory mapped then it is a relatively cheap operation, I think.

PR #266, opened 2020-12-18
Remove finish_observing 

danjampro:
Removes outdated code from `observatory` and `observing` state.

### DIFF
--- a/src/huntsman/pocs/states/huntsman/observing.py
+++ b/src/huntsman/pocs/states/huntsman/observing.py
@@ -17,7 +17,6 @@ def on_enter(event_data):
         wait_time = 0.
         while not all([event.is_set() for event in camera_events.values()]):
             pocs.logger.debug(f'Waiting for images: {wait_time} seconds.')
-
             sleep(WAIT_INTERVAL)
             wait_time += WAIT_INTERVAL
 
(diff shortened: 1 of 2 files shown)


src/huntsman/pocs/states/huntsman/observing.py
```diff
@@ -16,19 +16,15 @@ def on_enter(event_data):
 
         wait_time = 0.
         while not all([event.is_set() for event in camera_events.values()]):
```
JAAlvarado-Montes, line 18: Isn't it better using `panoptes.pocs.time.wait_for_events` here?

JAAlvarado-Montes (commented):
LGTM. Just a small comment about using `wait_for_events`.

src/huntsman/pocs/states/huntsman/observing.py
danjampro: @JAAlvarado-Montes Yes!

danjampro: Although I think we can leave this to a more generic cleanup of the `observe` function, which IMHO should/could be blocking.

JAAlvarado-Montes: OK, sounds good!

AnthonyHorton (approved):
LGTM.

Issue #267, opened 2020-12-18
The dome status is a method and not a property.


bazkiaei:
The dome status is a method while it should be a property. Currently we do `dome.status()` to get the dome status, after making it a property, it should change to `dome.status`.

PR #268, opened 2020-12-18
Making the dome status into a property instead of a method

bazkiaei:
The goal of this PR is to make the `dome.status()` a property instead of a method.
#267

### DIFF
--- a/scripts/byobu_startup.py
+++ b/scripts/byobu_startup.py
@@ -38,7 +38,7 @@ DOME_SHUTTER_STARTUP = ["from huntsman.pocs.dome.musca import HuntsmanDome",
                         "from huntsman.pocs.utils import load_config",
                         "config = load_config()",
                         "dome = HuntsmanDome(config=config)",
-                        "dome.status()",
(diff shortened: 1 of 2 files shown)


wtgee (approved):
This looks good in terms of getting some things working. Because it's reading a dictionary now it's relying on that dictionary being updated somewhere else. Might need something similar to an `_update_status` method?

bazkiaei:
> This looks good in terms of getting some things working. Because it's reading a dictionary now it's relying on that dictionary being updated somewhere else. Might need something similar to an `_update_status` method?

@wtgee Something like this?:

```
    @property
    def status(self):
        """A dictionary containing all status info for dome.

        TODO: Add other info (e.g. dome moving)
        TODO: Auto-update this every ~60s
        """
        status = self._update_status()
        return status

    def _update_status(self):
        if self._status_timer.expired():
            self._status = self._get_shutter_status_dict()
            self._status_timer = CountdownTimer(self._status_delay)
        return self._status

PR #274, opened 2020-12-27
Move all filterwheels to the luminance filter before focusing

JAAlvarado-Montes:
Hi everyone,

This PR adds a new private method that can be used to move ALL the filterwheels to a given filter. 

This new function is used now inside `autofocus_cameras` to move all the filterwheels to the `luminance` filter before focusing, which is needed during the `coarse_focusing` and `focusing` states. Also, this new function has replaced a piece of repeated code inside the `take_dark_images` method to move all the filterwheels to the `blank` position.

Please let me know any suggestion or comment. Thanks!

Cheers,
Jaime A.

### DIFF
--- a/conf_files/huntsman.yaml
+++ b/conf_files/huntsman.yaml
@@ -71,7 +71,7 @@ flat_fields:
   min_exptime: 0.0001
 
 darks:
-  exposure_times: [10, 30, 60, 300]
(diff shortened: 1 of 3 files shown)


lspitler (approved):
LGTM.

PR #276, opened 2021-01-06
Various fixes: Focusing, exposure error handling and tests

danjampro:
- Sets up tests so the coarse focusing code actually gets tested. Config modified accordingly.
- Use `concurrent.futures` to handle errors in `camera.pyro.client`. Added some tests.
- Actually return the move event from the pyro FW.
- Move testing config to `conf_files`


conf_files/testing.yaml
```diff
@@ -84,6 +84,7 @@ focusing:
     frequency: 2
     frequency_unit: hour
     timeout: 600
+    filter_name: one
```
lspitler, line 87: "one" doesn't seem like a filter name. can you rename this parameter?

lspitler (approved):
LGTM - one suggestion that might not be relevant

PR #277, opened 2021-01-06
Fw driver bug workaround

danjampro:
Not pretty but it works.


src/huntsman/pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,32 @@
...
+    def _move_to(self, position):
+
+        # Implement workaround for EFW library problem where cannot move to next-lowest position
+        # The workaround is to move to a different position before going to that position
+        # TODO: Remove this code when EFW library gets fixed!
+        n_filters = len(self._filter_names)
+        new_pos_int = self._parse_position(position)
+        diff = new_pos_int - self.position
+        if diff in (1, 1-n_filters):  # Problem only when this is True
+            for temp_pos in range(1, n_filters+1):  # Find a temporary position to move to first
```
JAAlvarado-Montes: add spaces for pep8 `n_filters + 1`

src/huntsman/pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def _move_to(self, position):
+
+        # Implement workaround for EFW library problem where cannot move to next-lowest position
+        # The workaround is to move to a different position before going to that position
+        # TODO: Remove this code when EFW library gets fixed!
+        n_filters = len(self._filter_names)
+        new_pos_int = self._parse_position(position)
+        diff = new_pos_int - self.position
+        if diff in (1, 1-n_filters):  # Problem only when this is True
```
JAAlvarado-Montes: Pep8: `1 - n_filters`

JAAlvarado-Montes (changes requested):

src/huntsman/pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def _move_to(self, position):
+
+        # Implement workaround for EFW library problem where cannot move to next-lowest position
+        # The workaround is to move to a different position before going to that position
+        # TODO: Remove this code when EFW library gets fixed!
+        n_filters = len(self._filter_names)
```
wtgee: Just an aside, but seems like this should be a class property already.

src/huntsman/pocs/filterwheel/zwo.py
wtgee: What are the possible filter positions? This is check for `1` and some other negative number (`-7`), so I'm assuming the first and or last position. But is that zero based? And should it be the absolute value?

src/huntsman/pocs/filterwheel/zwo.py
```diff
@@ -0,0 +1,32 @@
...
+            for temp_pos in range(1, n_filters+1):  # Find a temporary position to move to first
+                if temp_pos not in (self.position, position):
+                    break
+            # Move to the temporary position
+            self._driver.set_position(filterwheel_ID=self._handle,
+                                      position=self._parse_position(temp_pos) - 1,
+                                      move_event=self._move_event,
+                                      timeout=self._timeout)
+
+        # Filterwheel class used 1 based position numbering
```
wtgee, line 45: As per note above, this could be clarified some I suppose.

wtgee (approved):
LGTM. Maybe some log messages to explain what's going on?

src/huntsman/pocs/filterwheel/zwo.py
danjampro: The problem is that the filterwheel can't rotate a full turn minus one position. i.e. it can't go from position 3 to 2 or from 5 to 4 without raising an error, which is in the EFW library (we told them about it but no response). The problem also exists if we go from 1 to 7 (i.e. the number of filters), in which case the difference is `1-n_filters`.

PR #280, opened 2021-01-07
Refactor flats

danjampro:
Refactor flats into their own class:

- Declutters the HuntsmanObservatory class
- Starts to use `concurrent.futures`
- Takes biases if none provided
- Modular design

As @wtgee points out, it's not the final solution by any means. But a lot of this new code can be adapted for that.


wtgee:
While I see what you are getting at with this PR, I would be a bit worried about the direction it's taking. You're moving a lot of `Observatory` level logic into a new type of class that represents logical steps. 
 
The goal was always to make the calibration observations more similar to normal observations, which would ideally bring the sort of non-blocking `concurrent.futures` work you have here to regular observations.

I'm not saying you shouldn't go this direction, but might be worth a discussion before you put too much more time into it.

danjampro:
> While I see what you are getting at with this PR, I would be a bit worried about the direction it's taking. You're moving a lot of `Observatory` level logic into a new type of class that represents logical steps.
> 
> The goal was always to make the calibration observations more similar to normal observations, which would ideally bring the sort of non-blocking `concurrent.futures` work you have here to regular observations.
> 
> I'm not saying you shouldn't go this direction, but might be worth a discussion before you put too much more time into it.

@wtgee I totally agree that the long term direction should be to absorb the calibs into the scheduler as their own observations. But I don't see this happening any time soon (will it?). But I think the new code here will be easy to reuse in a future flat field observation class. In any case, the current code is a bit of a mess.

conf_files/testing.yaml
```diff
@@ -94,10 +94,11 @@ flat_fields:
     - deux
     - drei
     - quattro
-  required_exposures: 1  # Per filter per camera
-  max_attempts: 1  # Per filter per camera
-  max_exptime: 1 s
-  min_exptime: 0.0001 s
+  autoflats:
+    required_exposures: 1  # Per filter per camera
```
lspitler: SHould this be larger?

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,428 @@
...
+from astropy import units as u
+from astropy.io import fits
+
+from panoptes.utils.time import current_time, wait_for_events
+from panoptes.utils import get_quantity_value
+from huntsman.pocs.utils.logger import logger as LOGGER
+
+
+class AutoFlatFieldSequence():
+    """ Class to facilitate flat fields with automatic exposure time updates.
```
lspitler, line 18:
```suggestion
    """ Class to facilitate taking flat fields with automatic exposure time determination using current sky brightness. Works both for morning and evening twilights or even dome flats.
```

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+
+    def __init__(self, cameras, observation, initial_exposure_times=1*u.second,
+                 timeout=10*u.second, min_exptime=0.0001*u.second, max_exptime=60*u.second,
+                 max_attempts=10, required_exposures=5, target_scaling=0.16, scaling_tolerance=0.05,
```
lspitler, line 23: 0.16 seems low.

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,428 @@
...
+                 logger=None, safety_func=None, biases=None):
+        """
+        Args:
```
lspitler, line 26: Suggest adding an arg to specify the image slice where statistics are computed. Full frame stats computation is a waste of time and for our strongly vignetted lenses, not really going to be very meaningful.

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,428 @@
...
+            camera_name (str): The name of the camera.
+            elapsed_time (astropy.Quantity): The time between the previous exposure and now.
+        Returns:
+            astropy.Quantity: The next exposure time.
+        """
+        # Get data for specific camera
+        previous_exptime = self._exptimes[camera_name][-1]
+        target_counts = self._target_counts[camera_name]
+        average_counts = self._previous_counts[camera_name][-1]
```
lspitler:
```suggestion
        previous_counts = self._previous_counts[camera_name][-1]
```

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,428 @@
...
+
+        # Calculate next exptime
+        exptime = previous_exptime * (target_counts / average_counts)
+        sky_factor = 2.0 ** (elapsed_time / 180.0)
+        if self.past_midnight:
```
lspitler: don't see where `past_midnight` is set. I guess elsewhere?

lspitler: See note in slack - can this algorithm be modified to not do the midnight check, but instead detect if sky is falling or rising automatically and set the next exptime accordingly? This might mean it can be used for day or dome flats.

src/huntsman/pocs/autoflats.py
```diff
@@ -0,0 +1,448 @@
...
+
+        # Wait for exposures to complete
+        wait_for_futures(futures)
+
+        # Read the biases
+        self._biases = {}
+        for cam_name, filename in filenames.items():
+            self._biases[cam_name] = self._load_fits_data(filename)
+
+    def _take_darks(self, **kwargs):
```
JAAlvarado-Montes, line 273: Why having another function to take darks when there's already one (i.e. `take_dark_images` in `huntsman.pocs.observatory`)? Is it not better just importing that function here in `autoflats.py` and use it as it is intended to be used? Otherwise that function would only be used within the `taking_darks` state and this function `_take_darks` would just be duplicated code.

JAAlvarado-Montes (commented):
I think `take_dark_images` from `huntsman.pocs.observatory` should be used instead of `_take_darks`, or alternatively `take_dark_images` might be used inside `_take_darks`. As it is now I think it is kind of duplicated code.

src/huntsman/pocs/autoflats.py
wtgee: That's a bit too long for opening line but could be moved into docstring.

PR #281, opened 2021-01-07
Fix bug move all fws

JAAlvarado-Montes:
Hi,

This small PR removes one line that caused an error if the filterwheels were already in the requested filter.

Cheers,
Jaime.


danjampro (approved):
@JAAlvarado-Montes Looks good, if you want to make it neater you could put the whole thing on one line, like `filterwheel_events[camera] = camera.filterwheel...`

PR #284, opened 2021-01-11
Add option to skip solve WCS during analyzing_image

lspitler:
Solving for WCS may not be needed in all cases, so adding a configuration option to disable it. By default it will still happen.

Simply added an option `observations.analyize_recent_offset`. When set to False (not the default), it will skip the WCS solve field step after an image is taking.


JAAlvarado-Montes (approved):
LGTM!

PR #285, opened 2021-01-11
mount targets in docker

danjampro:
Small change to mount the full config directory inside the pocs container. Modifications to the config files on the control computer will be reflected inside the docker image.


JAAlvarado-Montes (approved):
LGTM!

PR #287, opened 2021-01-11
Move to given altaz coordinates for coarse focus if twilight flats are skipped

JAAlvarado-Montes:
Hi,

This PR fixes a small bug that occurs when the system is started after twilight flat fielding. That is, if POCS is started after flats, the mount would still be in the `parked` position and `coarse_focusing` is triggered with the telescope pointing at the floor. I have added here a conditional to move the telescope to the target when `twilight_flat_fielding` doesn't occur, so that we can coarse focus all the cameras pointing at the sky.

Cheers,
Jaime A.

### DIFF
--- a/conf_files/testing.yaml
+++ b/conf_files/testing.yaml
@@ -85,8 +85,8 @@ focusing:
     frequency_unit: hour
     timeout: 600
     filter_name: one
-    dummy_ra: 300
-    dummy_dec: -60
+    alt: 70  # degrees
+    az: 120  # degrees
 
 flat_fields:
   alt: 70  # degrees
--- a/conf_files/testing.yaml
+++ b/conf_files/testing.yaml
@@ -85,6 +85,8 @@ focusing:
     frequency_unit: hour
     timeout: 600
     filter_name: one
+    dummy_ra: 300
+    dummy_dec: -60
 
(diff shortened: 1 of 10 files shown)


danjampro (commented):
@JAAlvarado-Montes Thanks for pointing this out. But can we be sure that the target will be set at this point? i.e. to where will it slew?

JAAlvarado-Montes:
You're right @danjampro, we need to schedule the target at first. Not sure then if I need to trigger the `scheduling` state before doing this? or would it be better just putting some dummy RADEC coordinates? As what we'd need to do the coarse focus is any field and not necessarily the target? Which is somewhat similar to what we do with the flats.

conf_files/testing.yaml
```diff
@@ -85,6 +85,8 @@ focusing:
     frequency_unit: hour
     timeout: 600
     filter_name: one
+    dummy_ra: 300
```
danjampro: It might be better to specify this as an `alt-az` for now. I think you can drop the "dummy" in the variable name as its not really a dummy. So just `alt` and `az` are fine. You could even just make it use the flat field alt / az that is already in the config, but up to you.

danjampro (approved):
LGTM

PR #288, opened 2021-01-11
Improve dome status reliability

danjampro:
- Increase command delay from 0.5 to 1.0 s (now configurable)
- Implement configurable number of retries for valid status before raising an error
- Set `_status_timer` to None initially to prevent status being an empty dict if called soon after initialisation


fergusL (approved):
LGTM

PR #296, opened 2021-01-12
Get camera client timeout from proxy

danjampro:
At the moment there is no way to set  `camera.client._timeout` easily from the config. Solution is to set it to whatever the proxy has because that comes from the config.


JAAlvarado-Montes (approved):
LGTM!

wtgee:
Why not just query the config server directly? That's why it's on the network, so that remote devices can access it.

PR #297, opened 2021-01-12
Wait for config server and docker memory

danjampro:
- Moves the wait for config block into the pyro service function, allowing cameras to properly register with the name server after waiting for the config server to start (tested on Huntsman)
- Adds hard and soft memory limits to the docker container. Hopefully this will help some of the memory issues.

### DIFF
--- a/src/huntsman/pocs/utils/pyro/service.py
+++ b/src/huntsman/pocs/utils/pyro/service.py
@@ -1,12 +1,9 @@
-from multiprocessing import Process
-
-import Pyro5.errors
+import time
 from Pyro5.api import Daemon as PyroDaemon
 
-from panoptes.utils.config.client import set_config
 from panoptes.utils.library import load_module
+from panoptes.utils.config.client import get_config, server_is_running
 
-from huntsman.pocs.utils.config import get_config
 from huntsman.pocs.utils.logger import logger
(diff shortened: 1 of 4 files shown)


JAAlvarado-Montes (approved):
LGTM!

src/huntsman/pocs/utils/pyro/service.py
```diff
@@ -37,6 +34,10 @@ def pyro_service(service_class=None,
             service was started via `autostart=True`, the function will block until terminated,
             but still return the completed process.
     """
+    while not server_is_running():
+        logger.debug("Waiting for config server to start up before starting pyro service.")
+        time.sleep(5)
+
```
wtgee, line 40: This can hang forever.

I made the suggestion to use `wait-for-it` before the service starts, rather than part of the service itself.

https://github.com/AstroHuntsman/huntsman-pocs/issues/290#issuecomment-758152699

danjampro: @wtgee Thanks for the comment. Yes, it can hang forever, but I sort of want it to if the config server isn't running. Not sure what the advantages are of using `wait-for-it`? I'm trying to keep the shell scripts as minimal as possible, and we probably want to wait for the config server before starting pyro services in general.

PR #306, opened 2021-01-14
Using wrappers read and write from panopties.utils.rs232 in the HuntsmanDome class

JAAlvarado-Montes:
Hi,

This small PR replaces `self.serial.ser.write()` with `self.serial.write()` and `self.serial.ser.readlines()` with `self.serial.read()` from panoplies.utils.rs232 in order to have better control of error checking.

Please let me know any comment or suggestion.

Cheers,
Jaime A.

### DIFF
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -1,10 +1,13 @@
 # based upon @jamessynge's astrohaven.py code in POCS
-
 import threading
 import time
 
-from panoptes.pocs.dome.abstract_serial_dome import AbstractSerialDome
+import astropy.units as u
+
 from panoptes.utils import CountdownTimer
+from panoptes.utils import error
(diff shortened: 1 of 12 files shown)


src/huntsman/pocs/dome/musca.py
```diff
@@ -233,14 +233,14 @@ def _write_musca(self, cmd, log_message=None):
         """Write command to serial bluetooth device musca."""
         if log_message is not None:
             self.logger.info(log_message)
-        self.serial.ser.write('{}\n'.format(cmd).encode())
+        self.serial.write('{}\n'.format(cmd).encode())
```
danjampro: @JAAlvarado-Montes can you turn these into f-strings, i.e. "f{cmd}\n".encode()?

danjampro (approved):
LGTM, see comment about string formatting

PR #307, opened 2021-01-14
Fix bug dome

JAAlvarado-Montes:
Fix bug in the _write_musca private method when writing a command.


danjampro (approved):
LGTM, provided we still need the "\n"

PR #308, opened 2021-01-14
Fix bug dome removing .decode() from already decoded string

JAAlvarado-Montes:
This small PR removes .decode() from already decoded string.


danjampro (approved):
LGTM!

PR #309, opened 2021-01-14
Image archiving

danjampro:
Adds a new class that moves files from the images directory to the archive directory after they have been tracked for a certain amount of time. It runs as a docker service.


docker/docker-compose.yaml
```diff
@@ -19,6 +19,12 @@ volumes:
       type: none
       device: ${PANDIR}/huntsman-config/conf_files/pocs
       o: bind
+  archive:
```
lspitler, line 22:
```suggestion
  archiver:
```
to be consistent with the other yaml parameters?

conf_files/testing.yaml
```diff
@@ -115,3 +115,7 @@ pyro:
       - "huntsman"
       - "camera_service"
   devices: [ ]
+
+archiver:
+  sleep_interval: 1 s
```
lspitler, line 120: should this be longer in duration compared to `delay_interval` as per your recommendation below

src/huntsman/pocs/archive.py
```diff
@@ -0,0 +1,194 @@
...
+        """ Start archiving. """
+        self.logger.info("Starting archiving.")
+        self._stop = False
+        for thread in self._threads:
+            thread.start()
+
+    def stop(self, blocking=True):
+        """ Stop archiving.
+        Args:
+            blocking (bool, optional): If Trye (default), blocks until all threads have joined.
```
lspitler:
```suggestion
            blocking (bool, optional): If True (default), blocks until all threads have joined.
```

lspitler (approved):
LGTM. Have you tested it? Or is easier to test when merged?

conf_files/testing.yaml
danjampro: Could be, but its probably better for the tests to keep it like this, mainly because more things can go wrong in this case.

docker/docker-compose.yaml
danjampro: This corresponds to the archive directory and is not directly related to the archiver class

PR #310, opened 2021-01-15
Modify read musca to use self.serial.read() from panoptes-utils in a for loop

JAAlvarado-Montes:
This small PR modifies the `_read_musca` private method so that the status is returned faster. This was already tested on Huntsman and it's working well.

### DIFF
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -265,7 +265,7 @@ class HuntsmanDome(AbstractSerialDome):
            b'Door:Closed\r\n',
            b'Battery:\t 13.0671\r\n',
            b'Solar_A:\t 0.400871\r\n',
-           b'Switch:EM243A\r\n',
+           b'Switch:EM243A\r\n']
         """
         self._write_musca(Protocol.GET_STATUS)
         shutter_status_dict = {}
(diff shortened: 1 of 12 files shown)


src/huntsman/pocs/dome/musca.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, command_delay=1, max_status_attempts=3, *args, **kwargs):
 
         self._status = dict()
         self._status_delay = 5  # seconds
+        self._status_lines = 6
```
wtgee: We have a `Protocol.VALID_DEVICE` that specifies what keys are in the status call. We should use the length of that (+1) to query for the status so we don't have to track multiple things. See below.
```suggestion
```

src/huntsman/pocs/dome/musca.py
```diff
@@ -267,17 +260,17 @@ def _get_shutter_status_dict(self):
...
+           b'Battery:\t 13.0671\r\n',
+           b'Solar_A:\t 0.400871\r\n',
+           b'Switch:EM243A\r\n']
         """
         self._write_musca(Protocol.GET_STATUS)
-        shutter_status_list = self._read_musca()
         shutter_status_dict = {}
-        for shutter_status_item in shutter_status_list:
-            k, v = shutter_status_item.strip().decode().split(':')
+        for i in range(self._status_lines):
```
wtgee: From above:

```suggestion
        num_lines = len(Protocol.VALID_DEVICE)
        for i in range(num_lines + 1):  # Add one for the beginning 'Status' key
```

wtgee (approved):
LGTM! Thanks for working through that!

PR #311, opened 2021-01-15
Added explanatory comment to for loop in _get_shutter_status_dict

JAAlvarado-Montes:
Tiny PR. Adds a comment!

### DIFF
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -271,6 +271,8 @@ class HuntsmanDome(AbstractSerialDome):
         num_lines = len(Protocol.VALID_DEVICE)
         for i in range(num_lines + 1):  # Add one for the beginning 'Status' key
             k, v = self.serial.read().strip().split(':')
+            if k == 'Status':
+                continue
             if k == Protocol.SOLAR_ARRAY or k == Protocol.BATTERY:
                 v = float(v)
             shutter_status_dict[k] = v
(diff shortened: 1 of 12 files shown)


wtgee (approved):
👍🏼

src/huntsman/pocs/dome/musca.py
```diff
@@ -269,7 +269,7 @@ def _get_shutter_status_dict(self):
         self._write_musca(Protocol.GET_STATUS)
         shutter_status_dict = {}
         num_lines = len(Protocol.VALID_DEVICE)
-        for i in range(num_lines + 1):
+        for i in range(num_lines + 1):  # Add one for the beginning 'Status' key
```
wtgee, line 272: I probably should have mentioned this in the last PR, but it occurs to me now that we should probably not put the `Status` key in the `shutter_status_dict`.

So after getting `k, v` below, just a 

```
if k == "Status":
    continue
```

or some such.

wtgee (approved):

src/huntsman/pocs/dome/musca.py
```diff
@@ -269,8 +269,10 @@ def _get_shutter_status_dict(self):
...
             k, v = self.serial.read().strip().split(':')
+            if k == 'Status':
+                continue
             if k == Protocol.SOLAR_ARRAY or k == Protocol.BATTERY:
                 v = float(v)
             shutter_status_dict[k] = v
```
wtgee: Let's actually switch the logic a bit, which will also handle the coverage better:

```suggestion
            if k == Protocol.SOLAR_ARRAY or k == Protocol.BATTERY:
                v = float(v)
            if k != 'Status':
                shutter_status_dict[k] = v
```

Issue #304, opened 2021-01-13
Allow for failed exposures during flats and autofocus


danjampro:
We get failed exposures from time to time and need to have code that can deal with it. For example, the autofocus code could do a `try except` on each of its exposures, and only use the successful ones to calculate focus metrics.

wtgee:
We need to do this for regular observations as well, so much of the mechanics should be similar.

PR #314, opened 2021-01-15
Drop cameras from the observatory if the coarse focus fails.

wtgee:
This just checks which of the cameras successfully set the autofocus event and removes the ones that did not.

Related to #304

### DIFF
--- a/src/huntsman/pocs/states/huntsman/coarse_focusing.py
+++ b/src/huntsman/pocs/states/huntsman/coarse_focusing.py
@@ -1,7 +1,6 @@
-from panoptes.utils.time import wait_for_events, current_time
-
-from panoptes.utils import altaz_to_radec
-from panoptes.pocs.utils.location import create_location_from_config
+from panoptes.utils.time import CountdownTimer
+from panoptes.utils.time import current_time
+from panoptes.utils.utils import altaz_to_radec
 
 
 def on_enter(event_data):
@@ -14,8 +13,6 @@ def on_enter(event_data):
 
     # Setup information about earth location.
     pocs.say('Setting up location')
-    site_details = create_location_from_config()
-    earth_location = site_details['earth_location']
 
     # Set up the coarse focus config.
     coarse_focus_config = pocs.get_config('focusing.coarse')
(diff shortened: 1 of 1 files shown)


wtgee:
I should write a test...

src/huntsman/pocs/states/huntsman/coarse_focusing.py
```diff
@@ -1,7 +1,6 @@
-from panoptes.utils.time import wait_for_events, current_time
-
-from panoptes.utils import altaz_to_radec
-from panoptes.pocs.utils.location import create_location_from_config
+from panoptes.utils.time import CountdownTimer
```
JAAlvarado-Montes, line 1:
```suggestion

```

src/huntsman/pocs/states/huntsman/coarse_focusing.py
```diff
@@ -1,7 +1,6 @@
...
+from panoptes.utils.time import current_time
```
JAAlvarado-Montes, line 2:
```suggestion
from panoptes.utils.time import CountdownTimer, current_time
```

JAAlvarado-Montes (approved):
LGTM! Just a small suggestion, or maybe there's a reason I don't know for having the import like that :)

src/huntsman/pocs/states/huntsman/coarse_focusing.py
wtgee: Unfortunately we don't have a set style policy...

src/huntsman/pocs/states/huntsman/coarse_focusing.py
```diff
@@ -27,24 +24,48 @@ def on_enter(event_data):
...
+        # If overall timer has expired, check which cameras have not finished
+        # and remove them from the observatory.
+        if timer.expired():
+            pocs.logger.info(f'Autofocus timer expired, checking which cameras were successful')
+            for cam_name, focus_event in autofocus_events.items():
+                if not focus_event.is_set():
+                    pocs.logger.warning(f'Incomplete focus event on {cam_name}, removing')
+                    # TODO figure out why and also mark them for restart somehow.
+                    # See https://github.com/AstroHuntsman/huntsman-pocs/issues/304
+                    pocs.observatory.remove_camera(cam_name)
```
danjampro, line 58: I've seen problems before where changing the size of dictionaries while iterating over them can cause errors (e.g., https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error).

I'm not sure whether we will have that problem here? If this is already tested in `POCS` then it should be fine.

src/huntsman/pocs/states/huntsman/coarse_focusing.py
```diff
@@ -27,24 +24,48 @@ def on_enter(event_data):
...
-    wait_for_events(list(autofocus_events.values()), timeout=coarse_focus_timeout)
 
-    # Morning and not dark enough for observing...
-    if pocs.observatory.past_midnight and not pocs.is_dark(horizon='observe'):
-        pocs.next_state = 'twilight_flat_fielding'
-    else:
-        pocs.next_state = 'scheduling'
+    while True:
+        # Check if all cameras are finished and proceed if so.
+        if all([e.is_set() for e in autofocus_events.values()]):
```
danjampro, line 46: We are still using events for pyro autofocus, but at some stage we should switch to `concurrent.futures`, which implements a nice `wait` function (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) that will all us to use the timeout intrinsic to the focuser, which is a bit cleaner.

danjampro (approved):
Hi @wtgee looks good to me, provided that changing the dictionary size in the loop doesn't raise an error (see comment).

src/huntsman/pocs/states/huntsman/coarse_focusing.py
wtgee: This is iterating over a a dictionary of just camera names and `threading.Event` objects, so I think we should be okay.

PR #315, opened 2021-01-18
Musca dome robustness

danjampro:
- Add command lock to make thread safe
- Increase default command sleep to 10s


JAAlvarado-Montes (approved):
LGTM!

PR #316, opened 2021-01-18
Musca dome robustness

danjampro:
A few changes (tested on Huntsman) to make the dome more robust.

### DIFF
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -58,7 +58,7 @@ class HuntsmanDome(AbstractSerialDome):
     # V, so we don't open if less than this or CLose immediately if we go less than this
     MIN_OPERATING_VOLTAGE = 12.
 
-    def __init__(self, command_delay=1, max_status_attempts=10, move_timeout=60, sleep=120,
+    def __init__(self, command_delay=1, max_status_attempts=10, move_timeout=80, sleep=120,
                  *args, **kwargs):
         """
         Args:
@@ -122,8 +122,8 @@ class HuntsmanDome(AbstractSerialDome):
             for i in range(1, self._max_status_attempts + 1):
                 try:
                     status = self._get_status_dict()
-                    status["dome_thread"] = self._dome_thread.is_alive()
-                    status["keep_open"] = self._keep_open
+                    status["dome_thread_running"] = self._dome_thread.is_alive()
+                    status["keep_shutter_open"] = self._keep_open
                     return status
                 except Exception as err:
                     self.logger.warning(f"Retrying dome status: {i+1} of"
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -51,7 +51,6 @@ class HuntsmanDome(AbstractSerialDome):
 
     """
     LISTEN_TIMEOUT = 3  # Max number of seconds to wait for a response.
-    MOVE_TIMEOUT = 45  # Max number of seconds to run the door motors.
     MOVE_LISTEN_TIMEOUT = 0.1  # When moving, how long to wait for feedback.
     NUM_CLOSE_FEEDBACKS = 2  # Number of target_feedback bytes needed.
 
@@ -60,13 +59,14 @@ class HuntsmanDome(AbstractSerialDome):
     # V, so we don't open if less than this or CLose immediately if we go less than this
     MIN_OPERATING_VOLTAGE = 12.
 
-    def __init__(self, command_delay=1, max_status_attempts=10, *args, **kwargs):
+    def __init__(self, command_delay=1, max_status_attempts=10, move_timeout=60, *args, **kwargs):
         """
         Args:
             command_delay (float, optional): Wait this long in seconds before allowing next command
                 due to slow musica CPU. Default 10s.
             max_status_attempts (int, optional): If status fails, retry this many times before
                 raising a PanError. Default: 10.
+            move_timeout (u.Quantity, optional): The dome shutter movement timeout. Default 60s.
         """
         super().__init__(*args, **kwargs)
 
@@ -75,6 +75,7 @@ class HuntsmanDome(AbstractSerialDome):
         self._status = dict()
         self._close_event = threading.Event()
         self._command_delay = get_quantity_value(command_delay, u.second)
+        self._move_timeout = get_quantity_value(move_timeout, u.second)
         self._max_status_attempts = int(max_status_attempts)
 
         self._command_lock = Lock()  # Use a lock to make class thread-safe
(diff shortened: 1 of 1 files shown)


src/huntsman/pocs/dome/musca.py
```diff
@@ -3,11 +3,9 @@
 import time
 from threading import Lock
 
-import astropy.units as u
+from astropy import units as u
 
-from panoptes.utils import CountdownTimer
-from panoptes.utils import error
-from panoptes.utils import get_quantity_value
+from panoptes.utils import error, get_quantity_value
```
wtgee, line 7: We *really* need a set style policy...

src/huntsman/pocs/dome/musca.py
```diff
@@ -62,23 +59,23 @@ class HuntsmanDome(AbstractSerialDome):
     # V, so we don't open if less than this or CLose immediately if we go less than this
     MIN_OPERATING_VOLTAGE = 12.
 
-    def __init__(self, command_delay=10, max_status_attempts=10, *args, **kwargs):
+    def __init__(self, command_delay=1, max_status_attempts=10, move_timeout=60, *args, **kwargs):
         """
         Args:
             command_delay (float, optional): Wait this long in seconds before allowing next command
                 due to slow musica CPU. Default 10s.
```
wtgee:
```suggestion
                due to slow musca CPU. Default 1s.
```

src/huntsman/pocs/dome/musca.py
```diff
@@ -110,28 +107,14 @@ def status(self):
...
         TODO: Auto-update this every ~60s
         """
-        if self._status_timer is None:
-            update_status = True
-        else:
-            update_status = self._status_timer.expired()
-
-        if update_status:
+        with self._command_lock:  # Make status call thread-safe
             for i in range(1, self._max_status_attempts + 1):
```
wtgee:
```suggestion
            for i in range(self._max_status_attempts):
```

src/huntsman/pocs/dome/musca.py
```diff
@@ -62,23 +59,23 @@ class HuntsmanDome(AbstractSerialDome):
...
 
-    def __init__(self, command_delay=10, max_status_attempts=10, *args, **kwargs):
+    def __init__(self, command_delay=1, max_status_attempts=10, move_timeout=60, *args, **kwargs):
         """
         Args:
             command_delay (float, optional): Wait this long in seconds before allowing next command
                 due to slow musica CPU. Default 10s.
             max_status_attempts (int, optional): If status fails, retry this many times before
                 raising a PanError. Default: 10.
+            move_timeout (u.Quantity, optional): The dome shutter movement timeout. Default 60s.
```
wtgee: Seems like this should be `shutter_timeout`?

src/huntsman/pocs/dome/musca.py
```diff
@@ -234,13 +200,12 @@ def __del__(self):
 
     def _write_musca(self, cmd, log_message=None):
         """Wait for the command lock then write command to serial bluetooth device musca."""
-        with self._command_lock:
```
fergusL, line 237: why remove the lock here?

src/huntsman/pocs/dome/musca.py
```diff
@@ -150,8 +133,8 @@ def open(self):
                 'Dome shutter battery voltage too low to open: {!r}', v)
             return False
 
-        self._write_musca(Protocol.OPEN_DOME, 'Opening dome shutter')
-        time.sleep(HuntsmanDome.MOVE_TIMEOUT)
+        self._write_musca(Protocol.OPEN_DOME, 'Opening dome shutter.')
+        time.sleep(self._move_timeout)
```
wtgee: Is this just to wait until it's open?  Seems better to loop and check the `v == Protocol.OPEN` below and set a `max_wait` timer.

fergusL (commented):
Was there a reason to remove the lock on `write_musca`?

src/huntsman/pocs/dome/musca.py
wtgee: Yeah, the built in one second delay after write versus a lock seems odd.

wtgee (approved):
This isn't really my favorite, but I'll approve.

Ideally we should really try to find the musca source to see what is going on instead of building weird non-intuitive hacks into this code. But I don't think we can update musca until someone is on-site again 😦

src/huntsman/pocs/dome/musca.py
danjampro: I've moved the lock so we are blocking for the whole write/read for the `status` property. It is also implemented on write for `open` and `close` methods.

src/huntsman/pocs/dome/musca.py
```diff
@@ -141,106 +140,88 @@ def open(self):
         Boolean
             True if Opened, False if it did not Open.
         """
+        self._keep_open = True
```
wtgee: Should this be set after the two exception checks below?

src/huntsman/pocs/dome/musca.py
wtgee: You either need to change this as suggested (which makes more sense), or change the `{i+1}` below.

src/huntsman/pocs/dome/musca.py
```diff
@@ -104,34 +109,28 @@ def door_closed(self):
...
+        """A dictionary containing all status info for dome. """
+        with self._command_lock:  # Make status call thread-safe
             for i in range(1, self._max_status_attempts + 1):
+                try:
+                    status = self._get_status_dict()
+                    status["dome_thread_running"] = self._dome_thread.is_alive()
+                    status["keep_shutter_open"] = self._keep_open
+                    return status
+                except Exception as err:
+                    self.logger.warning(f"Retrying dome status: {i+1} of"
```
wtgee: See note above.

src/huntsman/pocs/dome/musca.py
```diff
@@ -104,34 +109,28 @@ def door_closed(self):
...
 
-        if update_status:
+    @property
+    def status(self):
+        """A dictionary containing all status info for dome. """
+        with self._command_lock:  # Make status call thread-safe
             for i in range(1, self._max_status_attempts + 1):
+                try:
+                    status = self._get_status_dict()
+                    status["dome_thread_running"] = self._dome_thread.is_alive()
```
wtgee: This is the dome status, so `dome_thread_running` isn't informative to the user (or is misinformative)

```suggestion
                    status["status_thread_running"] = self._dome_thread.is_alive()
```

src/huntsman/pocs/dome/musca.py
```diff
@@ -104,34 +109,28 @@ def door_closed(self):
...
+                    status["keep_shutter_open"] = self._keep_open
+                    return status
+                except Exception as err:
+                    self.logger.warning(f"Retrying dome status: {i+1} of"
+                                        f" {self._max_status_attempts}: {err}")
```
wtgee: Exception look better with the `!r` modifier.

```suggestion
                                        f" {self._max_status_attempts}: {err!r}")
```

wtgee (approved):

PR #317, opened 2021-01-18
Archiver tweaks

danjampro:
- Use `shutil` to move files because `os.rename` results in an `OSError`
- Raise error in script so docker knows to restart the service on error


bazkiaei (approved):
LGTM!

PR #319, opened 2021-01-18
Update run-camera-service.sh to include ssh reconnect

wtgee:
I think we should have the reconnect option here.

Really we should set up the cloud-init so that the Pis are mounting this as part of `fstab` rather than via a script.


danjampro:
LGTM

danjampro (approved):

PR #321, opened 2021-01-19
No local config file

danjampro:
- No loading / saving of a local config file
- Wait for config server for archiver

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -43,7 +43,7 @@ services:
       - config:/var/huntsman/huntsman-pocs/conf_files
       - panlog:/var/huntsman/logs
       - images:/var/huntsman/images
-    command: [ "panoptes-config-server --verbose run --config-file /var/huntsman/huntsman-pocs/conf_files/huntsman.yaml" ]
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yaml
```diff
@@ -43,7 +43,7 @@ services:
       - config:/var/huntsman/huntsman-pocs/conf_files
       - panlog:/var/huntsman/logs
       - images:/var/huntsman/images
-    command: [ "panoptes-config-server --verbose run --config-file /var/huntsman/huntsman-pocs/conf_files/huntsman.yaml" ]
+    command: [ "panoptes-config-server --verbose run --no-save-local --no-load-local --config-file /var/huntsman/huntsman-pocs/conf_files/huntsman.yaml" ]
```
wtgee, line 46: During our meeting yesterday we were confused as to the path for the config file and it's because we're mapping `huntsman-config` dir to `huntsman-pocs/conf_files/` inside the container.

Not at all related to this PR but thought I would mention here since we're looking at that line.

wtgee (approved):
LGTM

PR #322, opened 2021-01-19
Add POCS control service

wtgee:
* Adds a service that runs a no-op command. This should be used for controlling POCS via an ipython session.
* Add `json_store` volume so the storage persists outside the containers and so it can be written to for weather.


Physarah (approved):

JAAlvarado-Montes (approved):
LGTM!

Issue #318, opened 2021-01-18
Going to `ready` state while it is not safe to open the dome.


bazkiaei:
`pocs` goes to `sleeping` state when it is not safe to observe, e.g. there is no weather data available. But it just goes back to `ready` state and opens the dome.
@AnthonyHorton believes there might be some bugs related to weather safety checks, also `pocs.reset_observing_run()` in the `sleeping` state should handle it but it does not.

danjampro:
@bazkiaei I believe the "ready" state is tagged as "always_safe", so this is expected behaviour. This looks related to #299 ?

PR #328, opened 2021-01-20
State logic for opening dome

### LINKED ISSUE #323: Change dome open logic
When an observation is not ready to be observed (e.g. it is too close to the meridian) the dome will be closed and pocs goes to sleep. However if it is still safe to observe, it will reopen and repeat this process again in a loop of: 

observation not read -> sleep and close dome -> realise it is safe to observe -> open dome -> realise it’s too close to a flip -> sleep and close dome. 

Maybe we can add another level of 'sleep' such as 'idle' that occurs when the dome is open and ready to observe, but a target is not ready yet.

danjampro:
- Sleeping transitions to initialising, which opens the dome and prepares Huntsman
- Initialising transitions to ready, which can enter scheduling
- If nothing can be scheduled, go back to ready state (not initialising)

Closes #323 #318


src/huntsman/pocs/states/huntsman/initialising.py
```diff
@@ -0,0 +1,18 @@
...
+    pocs = event_data.model
+    pocs.next_state = 'parking'
+    pocs.say("Initialising Huntsman.")
+
+    pocs.observatory.prepare_cameras()
+    pocs.observatory.mount.unpark()
+    pocs.observatory.open_dome()
+
+    pocs.say("Initialisation complete.")
+    pocs.next_state == "ready"  # If not safe, the state machine goes to park automatically
```
wtgee: Should handle exceptions, at a minimum a wide net like:

```suggestion
    try:
        pocs.observatory.prepare_cameras()
        pocs.observatory.mount.unpark()
        pocs.observatory.open_dome()

        pocs.say("Initialisation complete.")
        pocs.next_state == "ready"  # If not safe, the state machine goes to park automatically
    except Exception as e:
        pocs.logger.warning(f'Cannot initialise {e!r}')
```

src/huntsman/pocs/states/huntsman/ready.py
```diff
@@ -1,11 +1,16 @@
...
+"""
+
+
 def on_enter(event_data):
     """
-    Once in the `ready` state our unit has been initialized successfully. We now
-    decide on the next state and ready the cameras if appropriate.
     """
     pocs = event_data.model
-    pocs.next_state = 'parking'
```
wtgee, line 7: By removing this is there still a fail-safe for if something goes wrong below?

wtgee (approved):
Looks good. Plan is to just to test live though?

Physarah (approved):
Happy to approve this but should we deploy and test it first?

Physarah:
It would be pretty easy to test by setting pocs time to the time last night when the error occurred

src/huntsman/pocs/states/huntsman/initialising.py
danjampro: @wtgee I think unhandled state errors should (ideally) be handled by the state machine, no?

src/huntsman/pocs/states/huntsman/ready.py
danjampro: See comment above. I think fail-safe should be handled by the state machine.

PR #329, opened 2021-01-20
Update POCS and panoptes-utils

wtgee:
* Python 3.8 for all tests and docker images.
* Update to `panoptes-pocs>=0.7.7` and `panoptes-utils>=0.2.31`
* Removed `conf_files` dir and contents in favor of `huntsman-config`.
* Pyro camera client uses `fits.open` with `output_verify` instead of `getdata`.
* Remove all explicit references to `PANUSER` or `PANDIR` (`PANLOG` is still pesky).
* `get_running_nameserver` will default to looking at `localhost`.
* Docker:
  * Built from `gcr.io/panoptes-exp/panoptes-pocs:develop`.
  * Installs `POCS` `develop` branch from github.
  * Simplifies image according to recent changes in POCS. **Note this changes directory structure inside the images. Volume mapping must be adjusted**.
  * Docker compose services started with `global` mode so they can't be started twice.
  * `docker/environment.yaml` for making conda installs inside docker images easier.
  * Removed explicit entrypoint. Default user is `1000` inside images.
* Testing:
  * all tests moved to `tests/` dir.
  * conftest starts config-server, nameserver and camera service. Removes `run_docker_tests.sh`.
  * config-server uses `tests/testing.yaml` for config.
  * Tests run from `tests/Dockerfile` using `tests/docker-compose.yml` using `scripts/testing/run_local_tests.sh`.
  * Simplified fixture creation (i.e. removed `config_with_simulated_stuff`).
  * Camera fixtures changed to `function` scope.
  * logs and coverage are output to local `logs` and `build` dir.
* Pyro service process creation function; used in `conftest.py`.
* Removing old `huntsman-shell` script.
* Misc cleanup.


wtgee:
Waiting on https://github.com/panoptes/POCS/pull/1068

docker/docker-compose.yaml
```diff
@@ -36,83 +36,82 @@ services:
...
-    depends_on:
-      - "pyro-name-server"
     environment:
-      HUNTSMAN_POCS: /var/huntsman
-      PANDIR: /var/panoptes
-      POCS: /var/panoptes/POCS
+      HUNTSMAN_POCS: /huntsman
       PANOPTES_CONFIG_PORT: 6563
-    command: [ "/var/panoptes/POCS/scripts/wait-for-it.sh localhost:6563 -- tail -f /dev/null" ]
+    command: [ "wait-for-it localhost:6563 -- tail -f /dev/null" ]
```
danjampro, line 91: What's this one all about?

wtgee: It just keeps the machine running so you can do an `exec` and connect.  But we can chat about this and if we want to start an `ipython` session by default.

src/huntsman/pocs/utils/pyro/service.py
```diff
@@ -1,23 +1,38 @@
...
+from Pyro5.api import Daemon as PyroDaemon
 from panoptes.utils.library import load_module
 from panoptes.utils.config.client import get_config, server_is_running
-
 from huntsman.pocs.utils.logger import logger
 from huntsman.pocs.utils.pyro.nameserver import get_running_nameserver
 from huntsman.pocs.utils.config import get_own_ip
 
 
+def pyro_service_process(auto_start=False, *args, **kwargs):
```
danjampro, line 12: We seemed to be hitting weird `Segmentation faults` on the pis when running pyro services inside Processes. Not sure exactly what the reason was though, but removing code similar to this resolved the problem. Just an FYI in case we see the same thing again!

wtgee: But it should be in a docker container on the pi right? Not sure how the Pi would know the different. But yeah, odd.

danjampro (approved):
LGTM! I'm in favour of merging and then testing on Huntsman.

src/huntsman/pocs/camera/pyro/client.py
```diff
@@ -1,23 +1,18 @@
...
-
+from panoptes.utils.time import CountdownTimer
+from panoptes.utils.utils import get_quantity_value
 from panoptes.pocs.camera import AbstractCamera
-
 from huntsman.pocs.filterwheel.pyro import FilterWheel as PyroFilterWheel
 from huntsman.pocs.focuser.pyro import Focuser as PyroFocuser
 from huntsman.pocs.utils.logger import logger
 from huntsman.pocs.utils.pyro.event import RemoteEvent
-from huntsman.pocs.utils.pyro import serializers  # Required to set up the custom (de)serializers
```
danjampro, line 20: Why was this removed?

PR #331, opened 2021-01-21
Refactor flats

danjampro:
- Flats no longer take darks - this should be handled by the "taking_darks" state
- Cameras that raise errors or timeout during exposures are now dropped from the observatory
- The flat field code should not hang while waiting for failed exposures


conf_files/testing.yaml
```diff
@@ -96,10 +96,11 @@ flat_fields:
...
     - quattro
-  required_exposures: 1  # Per filter per camera
-  max_attempts: 1  # Per filter per camera
-  max_exptime: 1 s
-  min_exptime: 0.0001 s
+  autoflats:
+    required_exposures: 2  # Per filter per camera
+    max_exposures: 2  # Per filter per camera
+    max_exptime: 1 s
+    min_exptime: 0.0001 s
```
Physarah, line 103: Maybe not relevant for this PR, but given the minimum exposure time for flats is 0.0001s, is it an easy config thing to change the sun altitude at which flats start, so that they can begin maybe 1 hour - 30 mins before sunset?

danjampro: yes, the flat field horizon is already configurable!

JAAlvarado-Montes:
@danjampro, the darks part could be triggered from the `twilight_flat_fielding` state. After taking flats, we can add something like this: `pocs.observatory.dark_images_required` to `twilight_flat_fielding`. Also, that would require to add the `dark_images_required` property to POCS (not sure if it has to be POCS or huntsman-pocs) so that it goes to the `taking_darks` state, take darks, and then move to `coarse_focusing`. Does that sound right?

conf_files/testing.yaml
lspitler: I *think* we might eventually do both "dayflats" and "twilight flats" everyday (and same thing in the morning?). That'll increase the likelihood of us obtaining some sort of flats on a particular day and will eventually give us a broader spectral range to fine-tune the flats for a particular science case. What we'd want to do is monitor how the spectrum of the sky during a clear day changes from well before sunset into astronomical twilight. It'll start blue and become redder as the sunset progresses. So eventually I think we'll have `min_daytime_extime` etc

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+        Returns:
+            float: The average counts.
+        """
+        try:
+            data = self._load_fits_data(filename)
+        except FileNotFoundError:
+            data = self._load_fits_data(filename + ".fz")
+
+        # Calculate average counts per pixel
+        average_counts, _, _ = sigma_clipped_stats(data - self._bias)
```
lspitler, line 143: If @AnthonyHorton is concerned about non-linearity, I think we want to *include* the bias in the stats calculation as it will push the pixels closer to a non-linearity part of their response function.

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+            self.logger.warning(f"Truncating exptime at maximum value.")
+            exptime = self._max_exptime
+
+        elif exptime <= self._min_exptime:
+            self.logger.warning(f"Truncating exptime at minimum value.")
+            exptime = self._min_exptime
+
+        return exptime
+
+    def _get_average_counts(self, filename, min_counts=1):
```
lspitler, line 131: Suggest adding the `stats_window_fraction=0.3` arg here and use it to save time during the stats computation, e.g:

```
x_dim = data.shape[1]
y_dim = data.shape[0]
windowed_data = data[int(np.round(stats_window_fraction * y_dim)):int(np.round((1 - stats_window_fraction) * y_dim)) , int(np.round(stats_window_fraction * x_dim)):int(np.round((1 - stats_window_fraction) * x_dim)) ]
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        try:
+            data = self._load_fits_data(filename)
+        except FileNotFoundError:
+            data = self._load_fits_data(filename + ".fz")
+
+        # Calculate average counts per pixel
+        average_counts, _, _ = sigma_clipped_stats(data - self._bias)
+        if average_counts < min_counts:
+            self.logger.warning('Clipping mean flat-field counts at minimum value: '
```
lspitler, line 145:
```suggestion
            self.logger.warning('Setting mean flat-field counts at minimum value: '
```

src/huntsman/pocs/utils/flats.py
lspitler: NM! @danjampro I see your much more elegant solution below! Much better.

lspitler (commented):
@danjampro only some comments. Probably should have @wtgee approve.

src/huntsman/pocs/utils/flats.py
danjampro: @lspitler maybe, but the bias count rate is small, and not including it here can mess up exposure times when the count rate is low (I think). Within the 5% tolerance I don't think it matters.

lspitler: Makes sense @danjampro

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+                target_counts=target_counts, counts_tolerance=counts_tolerance, bias=bias,
+                **kwargs)
+
+        # Loop until sequence has finished
+        self.logger.info(f"Starting flat field sequence for {len(cam_names)} cameras.")
+        while not all([s.is_finished for s in sequences.values()]):
+
+            if not safety_func():
+                self.logger.warning("Terminating flat fields because safety check failed.")
+                return
```
wtgee, line 553: Do you want to kill the observations here?

And this just relies on something else to properly park mount?

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+
+                # Start the exposure and get event
+                # TODO: Replace with concurrent.futures
+                events[cam_name] = camera.take_observation(
+                    observation, headers=headers, filename=filenames[cam_name],
+                    exptime=exptimes[cam_name])
+
+            # Wait for the exposures, dropping cameras that timeout
+            duration = get_quantity_value(max(exptimes.values()), u.second) + timeout
+            self._wait_for_camera_events(events, duration, remove=True)
```
wtgee: Some logging before and after this might be helpful.

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+
+            # Update the flat field sequences with new data
+            for cam_name in list(sequences.keys()):
+                if cam_name not in self.cameras.keys():  # Camera removed
```
wtgee: Should just be able to do this, which might be safer as `keys()` doesn't return a list directly.

```suggestion
                if cam_name not in self.cameras:  # Camera removed
```

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+                    del sequences[cam_name]
+                    continue
+                sequences[cam_name].update(filenames[cam_name], exptime=exptimes[cam_name],
```
wtgee:
```suggestion
                sequences[cam_name].update(filename=filenames[cam_name], exptime=exptimes[cam_name],
```

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+                                           time_start=start_times[cam_name])
+                # Log sequence status
+                status = sequences[cam_name].status
+                self.logger.info(f"Flat field status for {cam_name}: {status}")
+
+    def _autoflat_target_counts(self, cam_name, target_scaling, scaling_tolerance):
```
wtgee, line 598: This function seems like it could be useful other places than just buried in a private autoflat routine.

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+            bit_depth = 16
+
+        target_counts = int(target_scaling * 2 ** bit_depth)
+        counts_tolerance = int(scaling_tolerance * 2 ** bit_depth)
+
+        self.logger.debug(f"Target counts for {cam_name}: {target_counts}"
+                          f" ± {counts_tolerance}")
+        return target_counts, counts_tolerance
+
+    def _wait_for_camera_events(self, events, duration, remove=False, sleep=1):
```
wtgee: Something more explicit than just `remove` is probably a bit better. `remove_camera`?
```suggestion
    def _wait_for_camera_events(self, events, duration, remove_on_error=False, sleep=1):
```

src/huntsman/pocs/observatory.py
```diff
@@ -789,3 +514,127 @@ def _move_all_filterwheels_to(self, filter_name, camera_names=None):
...
+            duration (float): The total amount of time to wait for (should include exptime).
+            remove (bool, default False): If True, remove cameras that timeout. If False, raise
+                a TimeoutError instead.
+            sleep (float): Sleep this long between event checks. Default 1s.
+        """
+        timer = CountdownTimer(duration)
+        while not timer.expired():
+            if all([e.is_set() for e in events.values()]):
+                break
+            time.sleep(sleep)
```
wtgee, line 633: Is it still giving problems to do the `timer.sleep` instead of `time.sleep`?

src/huntsman/pocs/scheduler/observation.py
```diff
@@ -118,3 +118,22 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
         self._directory = os.path.join(self._image_dir, 'flats')
+
+    def get_exposure_filename(self, camera):
+        """ Get the exposure filename for a camera.
+        Args:
+            camera (Camera): A camera instance.
+        """
+        path = os.path.join(self.directory, camera.uid, self.seq_time)
+        filename = os.path.join(
+            path, f'flat_{self.current_exp_num:02d}.{camera.file_extension}')
```
wtgee: We could go above 99 in a sequence.

```suggestion
            path, f'flat_{self.current_exp_num:03d}.{camera.file_extension}')
```

src/huntsman/pocs/scheduler/observation.py
```diff
@@ -118,3 +118,22 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
 
         # Specify directory root for file storage
         self._directory = os.path.join(self._image_dir, 'flats')
+
+    def get_exposure_filename(self, camera):
```
wtgee, line 122: For flats?

```suggestion
    def get_flatfield_filename(self, camera):
```

src/huntsman/pocs/scheduler/observation.py
```diff
@@ -118,3 +118,22 @@ def __init__(self, position, pattern=dither.dice9, n_positions=9,
...
+            path, f'flat_{self.current_exp_num:02d}.{camera.file_extension}')
+        return filename
+
+    def get_bias_filename(self, camera):
+        """ Get the bias filename for a camera.
+        Args:
+            camera (Camera): A camera instance.
+        """
+        path = os.path.join(self.directory, camera.uid, self.seq_time)
+        filename = os.path.join(path, f'bias.{camera.file_extension}')
```
wtgee: Only ever a single bias?

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    def __init__(self, target_counts, counts_tolerance, initial_exposure_time=1*u.second,
+                 min_exptime=0.0001*u.second, max_exptime=60*u.second, max_exposures=10,
+                 required_exposures=5, cutout_size=300, bias=0, logger=None,):
+        """
+        Args:
+            target_counts (float): The target counts for each exposure.
+            counts_tolerance (float): The tolerance on target_counts in counts.
+                good flat field exposure.
```
wtgee: ?
```suggestion
            counts_tolerance (float): The tolerance on target_counts in counts for a
                good flat field exposure.
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+    @property
+    def is_finished(self):
+        """ Return True if the exposure sequence is finished, else False.
+        """
+        return self.status["is_finished"]
+
+    def update(self, filename, exptime, time_start):
+        """ Update the sequence data with the previous iteration.
+        Args:
+            average_counts (dict): The average counts.
```
wtgee:
```suggestion
            filename (str) The file to read the counts from.
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+            filename (str): The filename containing the data.
+            min_counts (float): The minimum count rate returned by this function. Can cause
+                problems if less than or equal to 0, so 1 (default) is a safe choice.
+        Returns:
+            float: The average counts.
+        """
+        try:
+            data = self._load_fits_data(filename)
+        except FileNotFoundError:
+            data = self._load_fits_data(filename + ".fz")
```
wtgee: You can be using the `panoptes.utils.images.fits.getdata` to automatically handle this. See below.

```suggestion
        data = self._load_fits_data(filename)
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
+from astropy.stats import sigma_clipped_stats
+from astropy import units as u
+from astropy.io import fits
```
wtgee: See notes below.

```suggestion
from panoptes.utils.images import fits as fits_utils
from panoptes.utils.images import crop_data
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+            filename (str): The FITS filename.
+            dtype (str or Type): The data type for the returned array.
+        Returns:
+            np.array: The exposure data.
+        """
+        data = fits.getdata(filename)
+        if self._cutout_size is not None:
+            x, y = data.shape[1]/2, data.shape[0]/2
+            data = Cutout2D(data, (x, y), size=self._cutout_size).data
+        return data.astype(dtype)
```
wtgee: This can be simplified using existing tools:

```suggestion
        data = fits_utils.getdata(filename).astype(dtype)
        if self._cutout_size is not None:
            data = crop_data(data, box_size=self._cutout_size)
            
        return data
```

src/huntsman/pocs/utils/flats.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        return abs(average_counts - self._target_counts) <= self._counts_tolerance
+
+    def _load_fits_data(self, filename, dtype="float32"):
+        """ Load FITS data, using a cutout if necessary.
+        Args:
+            filename (str): The FITS filename.
+            dtype (str or Type): The data type for the returned array.
+        Returns:
+            np.array: The exposure data.
```
wtgee:
```suggestion
            np.array: The exposure data clipped to _cutout_size and given in dtype.
```

wtgee (approved):
Mostly some style suggestions so @AnthonyHorton should still look at the specific questions regarding the linearity and bias.

danjampro:
> @danjampro, the darks part could be triggered from the `twilight_flat_fielding` state. After taking flats, we can add something like this: `pocs.observatory.dark_images_required` to `twilight_flat_fielding`. Also, that would require to add the `dark_images_required` property to POCS (not sure if it has to be POCS or huntsman-pocs) so that it goes to the `taking_darks` state, take darks, and then move to `coarse_focusing`. Does that sound right?

@JAAlvarado-Montes Good point. Following discussions with @AnthonyHorton we might move to scaling darks non-linearly with exposure time. So we will need to start gathering a) biases and b) darks with sequential exposure times. I think this should all happen in the dark state, and will be making a PR shortly.

src/huntsman/pocs/observatory.py
danjampro: @wtgee  Yes, it was my understanding that that sort of thing is the responsibility of the state machine?

danjampro:
> Mostly some style suggestions so @AnthonyHorton should still look at the specific questions regarding the linearity and bias.

@wtgee See discussion between @AnthonyHorton and myself on slack. I will be modifying the dark state shortly to address this.

src/huntsman/pocs/scheduler/observation.py
danjampro: This will be addressed in a future PR.

src/huntsman/pocs/observatory.py
danjampro: @wtgee  Yes... I would suggest it could be turned into a camera method.

src/huntsman/pocs/observatory.py
danjampro: Not sure, but this way we have less logger calls saying "sleeping for 1s"

PR #334, opened 2021-01-26
override initial_next_state default in POCS.run

danjampro:
- We need to go from sleeping to ready via the initialising state.
- We therefore need to change the default initial next state when running POCS.
- Added HuntsmanPOCS class with this override. The class will come in handy for overriding default state machine behaviour in future.


lspitler (approved):
@danjampro consider adding a unit test to catch this in the future.

PR #335, opened 2021-01-27
use trace level for archiver logging

danjampro:
Reduces the amount of archiver logs at the debug level.


JAAlvarado-Montes (approved):
LGTM!

PR #336, opened 2021-01-27
Switch to official docker buildx actions

danjampro:
- Current buildx actions are unreliable
- Official docker buildx setup is now available (https://github.com/docker/setup-buildx-action)


danjampro:
Merging after successful tests.

PR #339, opened 2021-02-01
moving aag-weather stuff into the main huntsman-pocs docker-compose file

fergusL:
copy n pasted' the aag-weather services into the main docker-compose file for huntsman-pocs.


wtgee:
@fergusL I made a note in slack about this but I don't think we want the AAG stuff as part of the POCS stuff because we (currently at least) restart all pocs services a lot.  Ideally we start up the weather as soon as the computer is on.

Note that you can use multiple compose files (e.g. `docker-compose -f huntsman-pocs.yaml -f aag-weather.yaml`).

It could still be used here but we should think it through first I suppose.  What is the motivation for combining them?

fergusL:
@wtgee, ah ok. No urgent reason, I think it was just a desire to move towards Dan's idea of only needing a single docker-comose.yml  on the control computer.

PR #351, opened 2021-02-11
Fix dome open logic in state machine

danjampro:
- Fix bug where dome would be opened *after* entering the state
- Override before_state method rather than add a new callback


danjampro:
Merging after testing on Huntsman.

PR #354, opened 2021-02-11
Stop mount tracking before slew

danjampro:
Stop mount tracking to stop potential for dome to be moving (to catch up with mount) when slew command is issued, resulting in a skyX error.


danjampro:
merging after testing on Huntsman

PR #372, opened 2021-03-01
Convert voltage to float

JAAlvarado-Montes:
This small PR fixes a bug where voltage if being left as string and not converted to float.

### DIFF
--- a/src/huntsman/pocs/dome/musca.py
+++ b/src/huntsman/pocs/dome/musca.py
@@ -128,7 +128,7 @@ class HuntsmanDome(AbstractSerialDome):
                     status["status_thread_running"] = self._dome_thread.is_alive()
                     status["keep_shutter_open"] = self._keep_open
 
-                    # Convert voltage and solar array to floats.
+                    # Convert voltage and solar array to floats
                     status[Protocol.BATTERY] = float(status[Protocol.BATTERY])
                     status[Protocol.SOLAR_ARRAY] = float(status[Protocol.SOLAR_ARRAY])
                     return status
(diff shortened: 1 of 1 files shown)


src/huntsman/pocs/dome/musca.py
```diff
@@ -113,7 +113,7 @@ def door_closed(self):
     @property
     def is_safe_to_open(self):
         v = self.status[Protocol.BATTERY]
-        if v < self.MIN_OPERATING_VOLTAGE:
+        if float(v) < self.MIN_OPERATING_VOLTAGE:
```
AnthonyHorton: This is the wrong place for this change. The function that gets the status for the dome should be converting number string to floats before it returns the status dict.

AnthonyHorton (changes requested):

src/huntsman/pocs/dome/musca.py
AnthonyHorton: In this function, should convert `status[Protocol.BATTERY]` and probably `status[Protocol.SOLAR_ARRAY]` to floats before returning the `status` dict :

https://github.com/AstroHuntsman/huntsman-pocs/blob/4949428c09eb27f08c19f516a35203e5dc7db711/src/huntsman/pocs/dome/musca.py#L122

JAAlvarado-Montes:
Is that better @AnthonyHorton ? Just addressed your comments :)

AnthonyHorton (approved):
Looks good to me. Hopefully there is no code elsewhere that expects these to be strings....

PR #373, opened 2021-03-01
make min_moon_sep adjustable via config

fergusL:


AnthonyHorton (approved):
LGTM (assuming it works).

PR #374, opened 2021-03-01
read in a min_moon_sep value from scheduler config

fergusL:
add the min_moon_sep value from config to the schedulers `common_properties` attribute so that it is used when `get_score()` is called on the MoonAvoidance constraint as part of `Scheduler.get_observation()`


wtgee (approved):
Should make a note to visit this in general for setting common properties but approving for now to get going.

Thanks!

PR #375, opened 2021-03-01
Allow custom moon separation

wtgee:
* Override the POCS `MoonAvoidance` constraint so we can pass the `min_moon_sep` properly.

This creates a default of `40` but also allows for setting in the config:

```
scheduler:
  type: dispatch
  fields_file: simple.yaml
  check_file: False
  constraints:
    min_moon_sep: 30
```


fergusL (approved):
LGTM

PR #378, opened 2021-03-02
Take flats in anti-solar direction

### LINKED ISSUE #376: Morning Flats should be taken point towards the anti-solar horizon
from http://adsabs.harvard.edu/pdf/1996PASP..108..944C
"
The null point for twilight flats is on the solar circle
near the zenith, offset toward the antisolar horizon by 0 to 40
deg. The penalty for misjudging the null point is around 1%
per degree gradient for every 15 deg off the null. The gradient can increase more rapidly than this near the sunward
horizon. That extra flat squeezed out of the brightest parts of
the dying twilight will probably have a very large gradient.
"

currently we are using {alt: 70 deg, az: 120 deg} as our flatfield pointing

danjampro:
Closes #376

### DIFF
--- a/src/huntsman/pocs/utils/flats.py
+++ b/src/huntsman/pocs/utils/flats.py
@@ -10,6 +10,28 @@ from panoptes.utils.utils import get_quantity_value
 from huntsman.pocs.utils.logger import logger as LOGGER
 
 
+from astropy.coordinates import get_sun
+from astropy.coordinates import AltAz
+
+
+def get_flat_field_altaz(location):
+    """ Return the alt/az opposite the Sun now at a given location.
+    Args:
+        location (astropy.coordinates.EarthLocation): The location.
+    Returns:
+        astropy.coordinates.AltAz: The alt/az opposite the Sun.
+    """
+    time_now = current_time()
+    frame = AltAz(obstime=time_now, location=location)
+    altaz_sun = get_sun(time_now).transform_to(frame)
+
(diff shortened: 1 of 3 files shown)


src/huntsman/pocs/utils/flats.py
```diff
@@ -10,6 +10,28 @@
...
+        location (astropy.coordinates.EarthLocation): The location.
+    Returns:
+        astropy.coordinates.AltAz: The alt/az opposite the Sun.
+    """
+    time_now = current_time()
+    frame = AltAz(obstime=time_now, location=location)
+    altaz_sun = get_sun(time_now).transform_to(frame)
+
+    # We want the anti-solar direction
+    alt = -1 * altaz_sun.alt  # Alt between -90 and 90 deg
```
fergusL: I don't think this is what we want for the alt.  "The null point for twilight flats is on the solar circle near the zenith, offset toward the antisolar horizon by 0 to 40 deg."

Until we have the time to do our investigation of this I think we should just stick with an alt of 70 degrees as before

fergusL (changes requested):
LGTM but I think we want a fixed alt of 70 degrees and only the azimuth changing to reflect the suns position

src/huntsman/pocs/utils/flats.py
danjampro: Makes sense!

fergusL (approved):
LGTM

PR #381, opened 2021-03-02
Fix ZWO camera reconnect 

danjampro:
Move some of the camera initialisation into the `connect` method.

### DIFF
--- a/src/huntsman/pocs/camera/zwo.py
+++ b/src/huntsman/pocs/camera/zwo.py
@@ -56,17 +56,16 @@ class Camera(AbstractSDKCamera):
         # default value is too low for their temperature resolution.
         self.temperature_tolerance = kwargs.get('temperature_tolerance', 0.6 * u.Celsius)
 
-        if gain:
-            self.gain = gain
+        self._gain = gain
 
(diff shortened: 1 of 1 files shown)


src/huntsman/pocs/camera/zwo.py
```diff
@@ -166,6 +165,11 @@ def connect(self):
         self._control_info = Camera._driver.get_control_caps(self._handle)
         self._info['control_info'] = self._control_info  # control info accessible via properties
         Camera._driver.disable_dark_subtract(self._handle)
+
+        if self._gain is not None:
+            self.gain = self._gain
+        self.image_type = self._image_type
```
AnthonyHorton, line 172: It's theoretically possible to get an `AttributeError` here. If `image_type` wasn't passed to `__init__` and the camera model doesn't support `RAW16` image type then `self._image_type` will not have been set. It can't happen with the type of cameras Huntsman has, but better to avoid the possibility anyway.

```suggestion
        with suppress(AttributeError):
            self.image_type = self._image_type
```

AnthonyHorton (approved):
LGTM.

PR #382, opened 2021-03-03
Improve musca dome (again again)

danjampro:
Not the PR that Huntsman deserves, but the PR Huntsman needs.

Tested on Huntsman.


danjampro:
Merging before tonight's run.

PR #384, opened 2021-03-03
tweak

fergusL:


src/huntsman/pocs/camera/zwo.py
```diff
@@ -55,8 +55,8 @@ def __init__(self,
         if image_type:
             self._image_type = image_type
         # Take monochrome 12 bit raw images by default, if we can
-        elif 'RAW16' in self.properties['supported_video_format']:
-            self._image_type = 'RAW16'
+        # elif 'RAW16' in self.properties['supported_video_format']:
+        #     self._image_type = 'RAW16'
         else:
             self._image_type = None
```
AnthonyHorton: No, we need it to default to `RAW16` if nothing else is specified, we just want to remove the check that RAW16 is a valid option before doing so, because that check needs to access the camera properties that aren't available at this point.

```suggestion
        else:
            self._image_type = 'RAW16'
```

AnthonyHorton (changes requested):
Needs to default to RAW16, not None.

AnthonyHorton (approved):
LGTM

PR #402, opened 2021-04-09
Optionally skip coarse focus

### LINKED ISSUE #400: Add option to skip states when running pocs
Something like:
`pocs.run(skip_darks=True, skip_flats=True, skip_coarse_focus=True)` etc

danjampro:
Adds ability to call `pocs.run` and skip initial coarse focus.

Closes #400

### DIFF
--- a/src/huntsman/pocs/core.py
+++ b/src/huntsman/pocs/core.py
@@ -9,7 +9,7 @@ class HuntsmanPOCS(POCS):
         self._dome_open_states = []
         super().__init__(*args, **kwargs)
 
-    def run(self, initial_next_state='starting', skip_coarse_focus=False, *args, **kwargs):
+    def run(self, initial_next_state='starting', initial_focus=True, *args, **kwargs):
         """ Override the default initial_next_state parameter from "ready" to "starting".
         This allows us to call pocs.run() as normal, without needing to specify the initial next
         state explicitly.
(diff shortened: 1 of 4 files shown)


lspitler (approved):

src/huntsman/pocs/core.py
```diff
@@ -8,15 +9,19 @@ def __init__(self, *args, **kwargs):
         self._dome_open_states = []
         super().__init__(*args, **kwargs)
 
-    def run(self, initial_next_state='starting', *args, **kwargs):
+    def run(self, initial_next_state='starting', skip_coarse_focus=False, *args, **kwargs):
```
wtgee: I'd suggest renaming it (and reverse the logic) so it's a bit more consistent from a user perspective. Either way the param should indicate it's only for the initial focus.

```suggestion
    def run(self, initial_next_state='starting', initial_focus=True, *args, **kwargs):
```

src/huntsman/pocs/core.py
```diff
@@ -8,15 +9,19 @@ def __init__(self, *args, **kwargs):
...
         state explicitly.
         Args:
             initial_next_state (str, optional): The first state the machine should move to from
                 the `sleeping` state, default `starting`.
+            skip_coarse_focus (bool, optional): If True, will skip the initial coarse focus.
+                Default False.
             *args, **kwargs: Parsed to POCS.run.
         """
+        if skip_coarse_focus:
+            self.observatory.last_coarse_focus_time = current_time()
```
wtgee: If you take suggestion up above for reversing logic then need:
```suggestion
        # Override last coarse focus time if not doing initial coarse focus.
        if initial_focus is False:
            self.observatory.last_coarse_focus_time = current_time()
```

wtgee (approved):
Minor suggestion otherwise LGTM.

PR #406, opened 2021-04-13
Observation revamp

### LINKED ISSUE #401: Add `DitheredSkyOffsetObservation`

danjampro:
Adds new abstract base classes for `Observation` and `Field` to make it easier to define fields consisting of multiple subfields (e.g. dithered fields, target + sky fields). 

Adds `CompoundField` and `CompoundObservation` to serve as base classes to observations with multiple fields. These can be configured from the config file since panoptes/POCS#1103. `CompoundObservation` objects have a `batch_size` parameter that allows us to take `batch_size` exposures for each subfield before moving on to the next one.

Adds `DitheredField` as a subclass of `CompoundField`, which is the input to `DitheredObservation` objects. It is possible to create a `CompoundField` consisting of multiple `DitheredFields` (e.g. for dithered target + dithered sky fields). See `test_scheduler.test_compound_dithered_observation` for example.

Eventually to be merged into `POCS`.

Closes #401


src/huntsman/pocs/scheduler/field.py
```diff
@@ -0,0 +1,142 @@
...
+            **kwargs: parsed to PanBase.__init__.
+        """
+        super().__init__(**kwargs)
+
+        self.equinox = equinox
+        self.frame = frame
+        self.name = name
+
+        # Prepare the field name
+        self._field_name = self.name.title().replace(' ', '').replace('-', '')
```
lspitler, line 28: Do you want to use underscores here instead?

lspitler: I can see why you don't, so fine with this convention as long as you've thought about it.

src/huntsman/pocs/scheduler/field.py
```diff
@@ -0,0 +1,142 @@
+from astroplan import FixedTarget
+from astropy.coordinates import SkyCoord
+
+from panoptes.utils.library import load_module
+from panoptes.pocs.base import PanBase
+from huntsman.pocs.utils import dither
+
+
+class AbstractField(PanBase):
```
lspitler, line 9: Do we wanna add the ability to change the exptime for each sub-field? Might be useful for dithering on/off a PSF reference bright star to get optimal PSF wing subtraction.

src/huntsman/pocs/scheduler/field.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def __init__(self, name, equinox="J2000", frame="icrs", **kwargs):
```
lspitler, line 11: Would it be too challenging to build-in mode where the target is moving? E.g. satellites or Elon's car?

tests/testing.yaml
```diff
@@ -32,7 +32,7 @@ directories:
   images: images
   mounts: /panoptes-pocs/resources/mounts
   resources: /panoptes-pocs/resources
-  targets: /panoptes-pocs/conf_files/targets
+  fields: /panoptes-pocs/conf_files/fields
```
lspitler, line 35: Reminder to update `huntsman-config` if needed?

lspitler (commented):
Looks good. THanks @danjampro. Mostly added some questions and points to ponder. Do agree @AnthonyHorton @wtgee should have a look.

src/huntsman/pocs/scheduler/field.py
danjampro: Would be nice but that doesn't work with the current `scheduler`. Definitely for future work though, as also useful for flat fields.

src/huntsman/pocs/scheduler/field.py
danjampro: That will require some thought but it should be possible. Not for this PR though!

src/huntsman/pocs/scheduler/field.py
danjampro: This is copy paste from `pocs`

src/huntsman/pocs/camera/zwo.py
```diff
@@ -177,6 +183,27 @@ def reconnect(self):
...
 
+    def take_exposure(self, defocused=False, *args, **kwargs):
+        """ Overrride class method to add defocusing offset. """
+        if defocused:
+            required_focus_offset = self._defocus_offset - self._current_defocus_offset
+            self.logger.debug(f"Applying focus offset of {self._defocus_offset} for defocused "
+                              "exposure.")
+        else:
+            self.logger.debug("Nullifying defocus offset.")
+            required_focus_offset = -self._current_defocus_offset
```
lspitler, line 194: as `defocused=False` won't this be run with every exposure? Do you wanna detect if `self._current_defocus_offset` has been set here before subtracting off a value?

src/huntsman/pocs/camera/zwo.py
```diff
@@ -177,6 +183,27 @@ def reconnect(self):
         self._reset_usb()
         return self.connect()
 
+    def take_exposure(self, defocused=False, *args, **kwargs):
+        """ Overrride class method to add defocusing offset. """
+        if defocused:
+            required_focus_offset = self._defocus_offset - self._current_defocus_offset
+            self.logger.debug(f"Applying focus offset of {self._defocus_offset} for defocused "
```
lspitler, line 190: I'd move this line down to where you actually move the focus.

Also the logs will be filled with a bunch of "Nullifying defocus offset" as `defocused=False`.

You could just detect one of 2 cases on line 197: `required_focus_offset` is positive or negative and you print either "Nullifying defocus offset" or "Applying focus offset of {self._defocus_offset} for defocused " depending on whether it is postive or negative?

fergusL (approved):
LGTM

src/huntsman/pocs/camera/zwo.py
danjampro: `self._current_defocus_offset` has a default value of 0. This line will always work regardless of how many times it is called.

src/huntsman/pocs/camera/zwo.py
danjampro: The required focus offset depends on whether the exposure is defocused or not, hence it is encapsulated within the `if` statement. The `nullifying focus offset` will appear in the debug log for every exposure - this is the sort of thing we want in the debug logs.

PR #411, opened 2021-04-20
Update Dockerfile

wtgee:
Fix paths for change in base docker image.  No longer need full paths to `conda`, `python`, `pip`, etc. Old full path also no longer works.

Depends on https://github.com/panoptes/POCS/pull/1105


danjampro (approved):
LGTM

PR #443, opened 2021-05-25
fix focus threshold bug

danjampro:


src/huntsman/pocs/utils/focus.py
```diff
@@ -13,16 +13,20 @@ class AutofocusSequence(PanBase):
...
         Args:
             position_min (int): The minimal focus position.
             position_max (int): The maximal focus position.
             position_step (int): The step in focus position.
             bit_depth (astropy.units.Quantity): The bit depth of the images.
-            mask_threshold (float, optional): The staturation masking threshold.
+            mask_threshold (float): The staturation masking threshold to apply to focus images.
+                Default: 0.9.
+            dark_mask_threshold (float, optional): The staturation masking threshold to eliminate
+                hot pixels from initial dark frame.
```
AnthonyHorton: The purpose of these might be clearer if they are called something like "saturation_mask_threshold" and "hot_pixel_mask_threshold".

src/huntsman/pocs/utils/focus.py
```diff
@@ -38,7 +42,7 @@ def __init__(self, position_min, position_max, position_step, bit_depth, mask_th
         self._position_max = int(position_max)
         self._position_step = int(position_step)
         self._bit_depth = bit_depth
-        self._mask_threshold = float(mask_threshold)
```
AnthonyHorton, line 41: You still need this line, don't you?

src/huntsman/pocs/utils/focus.py
```diff
@@ -109,15 +114,13 @@ def dark_image(self):
...
         Args:
             image (np.array): The dark image array.
         """
-        if self._mask is None:
-            self._mask = np.zeros(shape=image.shape, dtype="bool")
-
         self._dark_image = image.astype(self._image_dtype)
-        self._mask = np.logical_or(self._mask, self._mask_saturated(self._dark_image))
+        self._dark_mask = self._mask_saturated(self._dark_image,
+                                               threshold=self._dark_mask_threshold)
```
AnthonyHorton: Nit picking at naming again, but I think something like "saturated_pixel_mask" would be better than "dark_mask" at conveying the meaning/purpose of this thing.

AnthonyHorton (approved):
Looks good, I can only suggest some naming changes.

src/huntsman/pocs/utils/focus.py
danjampro: Yep, spotted that one!

Issue #444, opened 2021-05-27
Use AAT weather


danjampro:

wtgee:
This gets most of the way there. I don't know which of the `unknown_` columns are the actual safety readings, but I'm sure someone does.

Just need to add a `safe: True/False` value based off those columns.

```
import requests

url = 'http://aat-ops.anu.edu.au/met/metdata.dat'

columns = [
    'time',
    'ambient_temp',
    'dome_temp',
    'mirror_temp',
    'outside_dewpoint',
    'outside_humidity',    
    'pressure',
    'wind_speed_avg',
    'wind_gust_max',
    'wind_direction_avg',
    'unknown_3',
    'unknown_4',
    'unknown_5',
    'sky_ambient_diff_C',
    'sky_ambient_diff_error',    
    'boltwood_wetness_meter_maybe',
    'unknown_7',
    'unknown_8',
    'rain_since_9am',
    'sqm_brightness',
]

res = requests.get(url)
if res.ok:
    date, raw_data, _ = res.content.decode().split('\n')
    data = {name:value for name, value in zip(columns, raw_data.split('\t'))}
    data['date'] = date
```

which produces

```
{'time': '16:17:00',
 'ambient_temp': '7.9',
 'dome_temp': '5.9',
 'mirror_temp': '7.6',
 'outside_dewpoint': '-3.0',
 'outside_humidity': '46.2',
 'pressure': '667.1',
 'wind_speed_avg': ' 33',
 'wind_gust_max': ' 37',
 'wind_direction_avg': '180',
 'unknown_3': '0',
 'unknown_4': '0',
 'unknown_5': '0.6',
 'sky_ambient_diff_C': '-35.3',
 'sky_ambient_diff_error': '1.9',
 'boltwood_wetness_meter_maybe': '1858',
 'unknown_7': ' 0',
 'unknown_8': ' 0',
 'rain_since_9am': '0.0',
 'sqm_brightness': '0.00',
 'date': '" 05-27-2021."'}
```

fergusL:
For reference, here are the column names according to Steve Lee:

UT+10 i.e. local time, but NOT daylight saving time
Out_T Outside temperature in degrees C
In_T Inside dome temperature
Mir_T Temperature of the AAT primary mirror (measured a few millimetres above the actual surface)
Dew_O Dewpoint temperature of outside air
Hum_O% Humidity of outside air
Bar_mm Barometric pressure in mm of Mercury (for historical reasons)
[...]

PR #465, opened 2021-06-03
Use AATweather #444

fergusL:
- utility function to grab AAT weather 
- over-rides the pocs `is_weather_safe()` method so that it takes the result of the parent method and combines it with the AAT rain flag to determine if weather is safe

Potentially we might want to actually store the AAT data in a db somewhere, in which case I could make a script to query/store the data and add a new service to the docker-compose file to run this script.

Also looks like this change is going to break a bunch of tests that assume inserting a "safe" weather reading into the db will produce a corresponding safe result from the safety checks. Need a fixture or aat weather simulator


fergusL:
Added something so tests wouldn't fail. Tests passing locally but github actions is failing because it can't pull the pocs base image

wtgee:
@fergusL I'll take a closer look at this later, but we shouldn't be having any kind of `if` for tests in the actual code. If the db says `false` for safety then the weather is bad.

So you'll basically have to go the route of inserting into the database, which is what POCs is expecting. But it should be a simple `self.db.insert_current`. 

I'll look more later

src/huntsman/pocs/core.py
```diff
@@ -59,7 +59,7 @@ def after_state(self, event_data):
         """
         self.say(f"Finished with the {self.state} state. The next state is {self.next_state}.")
 
-    # Private methods
+    # private methods
```
danjampro: Just curious as to why this change? I have been commenting with capitals. Not important.

danjampro (commented):
@fergusL looks good. The thing that is missing seems to be how the state machine uses the safety data. I think all we need is a logical `and` to check whether weather is safe, e.g. something like:

`is_safe = aat_is_open & super().is_safe()`

in `core.HuntsmanPOCS`. It might also be good to add a config option to ignore AAT weather if we need to.

src/huntsman/pocs/core.py
fergusL: umm not sure, probably some auto formatting from my editor. Definitely not an intentional change

danjampro (approved):
Looks good! Do we need to keep the read_aat_weather script? If so then it should probably import things from `safety.py`. If not then it can be deleted.

Issue #466, opened 2021-06-04
add functionality to observe a target only once per night


lspitler:
The use-case is that we want to observe a bright star each night (or throughout the night) in order to get PSF wing information for use with the LSB science observations.

So we would need to add a flag (e.g. `observe_once`) to a particular target entry in the targets.yaml file. We need to also modify the scheduler to incorporate this type of condition.

There is a class that might be useful already:
https://github.com/panoptes/POCS/blob/86b68363488dfa0c55e649041add765228f48a9a/src/panoptes/pocs/scheduler/constraint.py#L162

PR #467, opened 2021-06-04
add observe_once constraint

lspitler:
This gives the user the ability to observe each target in the target list only once.

It is enabled via the config yaml file:

```
scheduler:
  constraints:
    observe_once: true
```

### DIFF
--- a/src/huntsman/pocs/utils/huntsman.py
+++ b/src/huntsman/pocs/utils/huntsman.py
@@ -66,7 +66,7 @@ def create_huntsman_scheduler(observer=None, logger=None, *args, **kwargs):
 
             # add global observe each target once constraint
             if get_config('constraints.observe_once', default=False):
-                constraints.append(AlreadyVisited())
+                constraints.insert(0, AlreadyVisited())
 
             # Create the Scheduler instance
             scheduler = module.Scheduler(observer, fields_file=fields_path, constraints=constraints,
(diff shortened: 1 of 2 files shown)


src/huntsman/pocs/utils/huntsman.py
```diff
@@ -16,6 +16,7 @@
 from huntsman.pocs.dome import create_dome_from_config
 from huntsman.pocs.core import HuntsmanPOCS
 from huntsman.pocs.scheduler.constraint import SunAvoidance
+from huntsman.pocs.scheduler.constraint import AlreadyVisited
```
wtgee: This is in the PANOPTES POCS. This should fix the failing test (hopefully)

```suggestion
from panoptes.pocs.scheduler.constraint import AlreadyVisited
```

src/huntsman/pocs/utils/huntsman.py
```diff
@@ -63,6 +64,10 @@ def create_huntsman_scheduler(observer=None, logger=None, *args, **kwargs):
                            HuntsmanMoonAvoidance(),
                            SunAvoidance()]
 
+            # add global observe each target once constraint
+            if get_config('constraints.observe_once', default=False):
+                constraints.append(AlreadyVisited())
```
wtgee: The scheduler will veto items in order, so it's best if the constraints can veto early in the process. If it's already been visited there is no reason to check, e.g. distance from the moon.

```suggestion
                constraints.insert(0, AlreadyVisited())
```

wtgee (approved):
Approving, but it needs to be fixed to pass tests.  Could probably also use a test of its own.

lspitler:
Related to #466 but that issue is for individual targets whereas the functionality captured here applies to all targets.

PR #531, opened 2021-08-27
_take_autoflats() determines initial exp time by whether it is AM or PM

fergusL:


danjampro (approved):
LGTM. At some point we should move this to a config item, but since there are a lot of places where we should do that let's skip it for now.

fergusL:
Ya, I have another pr in huntsman-config to add it to the autoflats config, so should just get passed in as a kwarg from the autoflat config here https://github.com/AstroHuntsman/huntsman-pocs/blob/db03d12d92d66eb1b4a7a88c63653e228b15f4d9/src/huntsman/pocs/observatory.py#L275

PR #541, opened 2021-10-08
Added util func for skymapper weather and query/store in db script

fergusL:
added a util that will return a skymapper weather data dict of form,

```
{'Date': '2021-10-08',
 'Time': '01:53:00',
 'Raining': False,
 'Cloudy': True,
 'Humidity': False,
 'Windy': False,
 'Hailing': False,
 'Excessive_Light': True,
 'Internal_Rain_Sensor': False,
 'External_Rain_Sensor': False,
 'Relative_Humidity': '48',
 'Wind_Speed_Minimum': '3.60',
 'Wind_Speed_Average': '4.06',
 'Wind_Speed_Maximum': '4.20',
 'Sky_Minus_Ambient': '-32.00',
 'Wind_Speed': '10.40'}
```

and a script that can be run to add this data to weather db periodically


fergusL:
- cleaned up functions that fetch aat/skymapper weather data
- added function/config for determining which sources to use and which parameters to base safety decisions on
- added generic script for logging alternate weather data
- updated huntsman.pocs.core to use alternate weather sources as specified in config (ie whether to use a source and what weather readings to use)
- updated testing config to ignore alt weather sources

scripts/read_alt_weather.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def main(source,
+         config,
+         store_result,
+         storage_dir,
+         read_delay,
+         verbose,
+         **kwargs):
+
```
danjampro, line 20: Could add a docstring here explaining what the function does and what the args are.

scripts/read_alt_weather.py
```diff
@@ -0,0 +1,68 @@
...
+            config = yaml.safe_load(f)
+            alt_weather_config = config.get('alt_weather_sources').get(source)
+
+    if alt_weather_config is None:
+        raise ValueError("The weather source config is None, please supply a config.")
+
+    while True:
+        try:
+            # load and call function for fetching alterant weather source data
+            data = load_module('huntsman.pocs.utils.weather.get_' + source)()
```
danjampro: Could take the `load_module` call outside of the loop

scripts/read_alt_weather.py
```diff
@@ -0,0 +1,68 @@
...
+            # determine if the weather reading indicates safe conditions
+            data = determine_alt_weather_safety(data, alt_weather_config)
+            if verbose:
+                print(f'{data!r}')
```
danjampro: Can we use the logger instead?

src/huntsman/pocs/core.py
```diff
@@ -114,21 +112,51 @@ def is_weather_safe(self, **kwargs):
         if self._in_simulator('weather'):
             return True
         # if not in simulator mode, determine safety from huntsman weather data
-        is_safe = super().is_weather_safe(**kwargs)
+        is_safe_list = [super().is_weather_safe(**kwargs)]
+
+        # check alternate sources
+        alt_weather_config = self.get_config("alt_weather_sources", default=None)
+        sources_to_check = [s for s in alt_weather_config.keys()
```
danjampro: Think we can simplify this:
`[k for k, v in alt_weather_config.items() if v['use']]`

danjampro (approved):
LGTM! A few minor comments above.

PR #543, opened 2021-11-12
mutli filter obs changes

fergusL:
Created new `AbstractHuntsmanCamera` base class so that `observation.filter_name` can be overriden on a per camera basis. Also moved exposure time tuning from `CameraGroup` into `AbstractHuntsmanCamera` base calss.


src/huntsman/pocs/camera/camera.py
```diff
@@ -0,0 +1,187 @@
...
+                                                                         **kwargs)
+
+        # pop exptime from kwarg as its now in exptime
+        exptime = kwargs.pop('exptime', observation.exptime.value)
+
+        # Check if we need to tune the exposure time
+        # (should this go in setup_observation?)
+        with suppress(AttributeError):
+            if observation.target_exposure_scaling is not None:
+                exptime = tune_exposure_time(target=observation.target_exposure_scaling,
```
danjampro: I think the camera is the first argument to `tune_exposure_time`. Actually, now that we are overriding the `Camera` class I think we can add `tune_exposure_time` as a new `Camera` method. So we could do `self.tune_exposure_time(...`

src/huntsman/pocs/camera/camera.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        observation_event = threading.Event()
+        # Setup the observation
+        exptime, file_path, image_id, metadata = self._setup_observation(observation,
+                                                                         headers,
+                                                                         filename,
+                                                                         **kwargs)
+
+        # pop exptime from kwarg as its now in exptime
+        exptime = kwargs.pop('exptime', observation.exptime.value)
```
danjampro, line 45: `exptime = kwargs.pop('exptime', observation.exptime.to_value("s"))`

src/huntsman/pocs/camera/camera.py
```diff
@@ -0,0 +1,187 @@
...
+            name=f'Thread-{image_id}',
+            target=self.process_exposure,
+            args=(metadata, observation_event),
+            daemon=True)
+        t.start()
+
+        if blocking:
+            while not observation_event.is_set():
+                self.logger.trace('Waiting for observation event')
+                time.sleep(0.5)
```
danjampro: 0.5s is probably a little too high for us.

src/huntsman/pocs/camera/group.py
```diff
@@ -224,6 +181,23 @@ def autofocus(self, *args, **kwargs):
         cameras = {n: c for n, c in self.cameras.items() if c.has_focuser}
 
         def func(cam_name):
-            return cameras[cam_name].autofocus(*args, **kwargs)
+            # need to remove `filter_name` from kwargs so it only gets passed once
+            # to `cameras[cam_name].autofocus()`
+            filter_name = kwargs.pop('filter_name')
```
danjampro: Think this should be moved out of the inner function

src/huntsman/pocs/camera/group.py
```diff
@@ -224,6 +181,23 @@ def autofocus(self, *args, **kwargs):
...
+    # Private methods
+
+    def _get_focus_filter_name(self, camera_name, filter_name=None, coarse=False, *args, **kwargs):
+        """
+        """
+        if coarse or filter_name is None:
+            filter_name = self.get_config('focusing.coarse.filter_name')
+            return filter_name
+        elif isinstance(filter_name, abc.Mapping):
+            return filter_name[camera_name]
```
danjampro: Do we want to handle the potential `KeyError` here?

src/huntsman/pocs/scheduler/observation/base.py
```diff
@@ -214,6 +209,21 @@ def mark_exposure_complete(self):
...
+        """ Determine filter name based on camera if `filter_name` is dict,
+        otherwise just return `filter_name`.
+
+        Args:
+            camera_name (str): The name of the camera.
+        Returns:
+            str: The filter name.
+        """
+        # If filter names is a dict, use camera name as key
+        if isinstance(self.filter_name, abc.Mapping):
```
danjampro, line 224: Think we should log and raise a `PanError` here if there is a `KeyError` on the camera name

src/huntsman/pocs/camera/camera.py
danjampro: I think this is where the current test error is coming from. The argument name is (annoyingly) `seconds` instead of `exptime` here - that is why multiple `seconds` values get parsed into `take_exposure` (maybe)

danjampro (commented):
Looks good! Very close to finished. Hopefully the "seconds / exptime" switch will get the tests passing.

lspitler:
@danjampro looks like @fergusL needs a final review?

lspitler:
merging so @fergusL can test it tomorrow.

PR #547, opened 2022-04-05
Changes to bring HPOCS in line with how POCS handles `observing_event`

fergusL:
In POCS the `camera.take_observation()` method was changed to return  a `metadata` object instead of the `observation_event` it was before.

https://github.com/panoptes/POCS/blob/83ced36d0d6bd85af31a728246b47da470584d7d/src/panoptes/pocs/camera/camera.py#L460

Now `observation_event` has been made into a class attribute called `camera._is_observing_event` which can be interacted with via @property attribute `camera.is_observing`.

Changes in this PR are to bring HPOCS in line with the above chagnes to pocs but also required making some changes to the pyro camera client class and some of the code related to handling remote events through pyro.

### DIFF
--- a/src/huntsman/pocs/camera/camera.py
+++ b/src/huntsman/pocs/camera/camera.py
@@ -12,6 +12,7 @@ from astropy import units as u
 
 
 class AbstractHuntsmanCamera(AbstractCamera):
+    """Base class for Huntsman cameras"""
 
     def take_observation(self, observation, headers=None, filename=None, blocking=False, **kwargs):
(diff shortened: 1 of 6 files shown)


fergusL:
Hey @AnthonyHorton, if you have a second could you please take a look at the error I'm getting on the flat fields tests? There have been some changes in POCS with how the `Camera` observation event is implemented and I've tried to bring Hpocs in line with the POCS changes and make it work with the pyro interface but it is causing issues for parallel camera tasks...

PR #597, opened 2023-07-31
Adding upload capabilities to huntsman-pocs archiver

bazkiaei:


scripts/archive-images.py
```diff
@@ -4,7 +4,6 @@
 from huntsman.pocs.archive.archiver import Archiver
 
 if __name__ == "__main__":
-
```
lspitler: Please undo this as it serves no purpose.

scripts/sync_data_with_remote_server.py
```diff
@@ -0,0 +1,158 @@
...
+        archive_directory=archive_directory,
+        remote_host=remote_host,
+        username=username,
+        password=password,
+        port=port,
+    )
+    archiver.start()
+
+    # Monitor the archiver
+    # If it breaks we want to terminate the script so docker can restart the service
```
lspitler: I don't think docker will be doing any restarting of the service. This should run outside of docker.

scripts/sync_data_with_remote_server.py
```diff
@@ -0,0 +1,158 @@
...
+    while True:
```
lspitler: Can you add a `try` that gracefully handles a Keyboardinterrupt. What if this happens during a file transfer?

bazkiaei: The introduced class in this file looks for `fits` files and copy them to the destination and delete each file after copying is completed.

Therefore in a scenario like the one you mentioned, the file that its copying is not completed still exist on the control computer and after re-running the code, the code finds it and transfers it to the server. The code deletes the file when the copying is completed.

scripts/sync_data_with_remote_server.py
bazkiaei: I started working on this by playing with `archive-images.py` and this comment is left there from that file. I will remove it.

scripts/sync_data_with_remote_server.py
bazkiaei: Oh! I forgot to add the `try` part! will do that in the next commit when receive your comments.

bazkiaei:
Hey @fergusL ! This PR is ready to be reviewed!

fergusL (approved):
lgtm

